PR #1, opened 2024-08-29
feat(config): Validation and typing for configuration file

hubertdeng123:
This uses pydantic to implement a more rigorous structure for devservices configuration files

### DIFF
--- a/src/utils/config.py
+++ b/src/utils/config.py
@@ -1,11 +1,88 @@
 from __future__ import annotations
 
 import os
-from typing import Dict
+from typing import Dict, List, Optional, Union
 
 import yaml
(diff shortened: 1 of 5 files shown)


src/utils/config.py
```diff
@@ -1,11 +1,88 @@
 from __future__ import annotations
 
 import os
-from typing import Dict
+from typing import Dict, List, Optional, Union
```
hubertdeng123, line 4: @IanWoodard do we really want there to be separate lines for this? Seeing CI fail
```
-from typing import Dict, List, Optional, Union
+from typing import Dict
+from typing import List
+from typing import Optional
+from typing import Union
```

IanWoodard (approved):

PR #3, opened 2024-09-04
feat(services): Adding list-services command

IanWoodard:
Doing some general work to improve how we get information about available services to run:
* Adding list-services command
* Adding new logic for determining which services are available locally

### DIFF
--- a/src/commands/list_dependencies.py
+++ b/src/commands/list_dependencies.py
@@ -5,6 +5,9 @@ from argparse import ArgumentParser
 from argparse import Namespace
 
 from configs.service_config import load_service_config
+from services import find_matching_service
+from services import get_local_services
+from utils.devenv import get_coderoot
 
 
 def add_parser(subparsers: _SubParsersAction[ArgumentParser]) -> None:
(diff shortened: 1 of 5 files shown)


src/services.py
```diff
@@ -0,0 +1,41 @@
+from __future__ import annotations
+
+import os
+from typing import List
+
+from configs.service_config import load_service_config
+from pydantic import BaseModel
+
+
+class Service(BaseModel):
```
hubertdeng123, line 10: I thought about it a bit, and think we can use [dataclasses](https://docs.python.org/3/library/dataclasses.html) instead of pydantic. I'll be putting up a PR to change this given that pydantic is slower, and I don't see our data model getting large and complex to the point where we'd take a benefit there.

src/commands/list_dependencies.py
```diff
@@ -22,8 +25,23 @@ def add_parser(subparsers: _SubParsersAction[ArgumentParser]) -> None:
...
+    service_name = args.service_name
+
+    # Get all of the services installed locally
+    coderoot = get_coderoot()
+    local_services = get_local_services(coderoot)
+
+    repo_path = None
+    # If a service name is provided, try to find the matching service
+    if service_name is not None:
+        try:
```
hubertdeng123, line 37: What was your reasoning for moving logic for determining repo path outside of `load_service_config`? I believe it should live there instead, since cli commands should not be handling this logic. That way, every time a service name is not defined for a command, the logic to load config stays the consistent and it should not change.

PR #10, opened 2024-09-09
feat(cli): Add start and stop

hubertdeng123:
Adds start/stop basic cli commands that use the services in the default mode.

### DIFF
--- a/src/commands/start.py
+++ b/src/commands/start.py
@@ -1,27 +1,48 @@
 from __future__ import annotations
 
+import os
+import subprocess
 from argparse import _SubParsersAction
 from argparse import ArgumentParser
 from argparse import Namespace
 
+from constants import DEVSERVICES_DIR_NAME
+from constants import DOCKER_COMPOSE_FILE_NAME
 from utils.services import find_matching_service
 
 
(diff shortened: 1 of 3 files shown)


src/commands/stop.py
```diff
@@ -1,28 +1,47 @@
...
+        [
+            "docker",
+            "compose",
+            "-f",
+            service_config_file_path,
+            "down",
+        ]
+        + mode_dependencies,
+        check=False,
+    )
```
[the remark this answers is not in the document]
hubertdeng123: yeah you're right, we should use that

src/commands/start.py
```diff
@@ -30,19 +30,11 @@ def start(args: Namespace) -> None:
...
-            service_config_file_path,
-            "up",
-            "-d",
-        ]
-        + mode_dependencies,
-        check=False,
+    output = run_docker_compose_command(
+        f"-f {service_config_file_path} up -d {mode_dependencies}"
     )
+    print(output)
```
IanWoodard, line 40: What is the reasoning behind printing the output here?

hubertdeng123: ah my file wasn't saved locally, I removed it but it wasn't pushed. good catch

IanWoodard (approved):

PR #21, opened 2024-09-16
Adding status loader utility

IanWoodard:
Adding a basic status loader utility to make the CLI more user friendly. This utility spins as long-running commands process to show the user that work is being down behind the scenes, improving UX.

### DIFF
--- a/src/commands/start.py
+++ b/src/commands/start.py
@@ -9,6 +9,7 @@ from constants import DEVSERVICES_DIR_NAME
 from constants import DOCKER_COMPOSE_FILE_NAME
 from utils.docker_compose import run_docker_compose_command
 from utils.services import find_matching_service
+from utils.terminal import Status
 
 
 def add_parser(subparsers: _SubParsersAction[ArgumentParser]) -> None:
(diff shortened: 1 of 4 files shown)


src/commands/start.py
```diff
@@ -35,10 +36,11 @@ def start(args: Namespace) -> None:
...
         service.repo_path, DEVSERVICES_DIR_NAME, DOCKER_COMPOSE_FILE_NAME
     )
-    try:
-        run_docker_compose_command(
-            f"-f {service_config_file_path} up -d {mode_dependencies}"
-        )
-    except DockerComposeError as dce:
-        print(f"Failed to start {service.name}: {dce.stderr}")
-        exit(1)
+    with Status(f"Starting {service_name}", f"{service_name} started"):
```
hubertdeng123, line 39: If service_name is `None`, that will be printed here. Instead of that, we should use service.name

IanWoodard: Good catch!

hubertdeng123 (approved):

PR #28, opened 2024-09-17
feat(remove): Adding remove command

IanWoodard:
Adding basic remove command to remove volumes for services

### DIFF
--- a/src/commands/remove.py
+++ b/src/commands/remove.py
@@ -1,0 +1,49 @@
+from __future__ import annotations
+
+import os
+from argparse import _SubParsersAction
+from argparse import ArgumentParser
+from argparse import Namespace
+
+from constants import DEVSERVICES_DIR_NAME
+from constants import DOCKER_COMPOSE_FILE_NAME
+from exceptions import DockerComposeError
+from utils.console import Status
+from utils.docker_compose import run_docker_compose_command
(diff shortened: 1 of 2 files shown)


src/commands/remove.py
```diff
@@ -0,0 +1,49 @@
...
+from constants import DOCKER_COMPOSE_FILE_NAME
+from exceptions import DockerComposeError
+from utils.console import Status
+from utils.docker_compose import run_docker_compose_command
+from utils.services import find_matching_service
+
+
+def add_parser(subparsers: _SubParsersAction[ArgumentParser]) -> None:
+    parser = subparsers.add_parser("remove", help="Remove a volume for a service")
+    parser.add_argument(
```
hubertdeng123, line 18: hmmm if I'm reading this correctly this would read as `devservices remove snuba clickhouse` to remove the clickhouse volume right? Seems a bit odd...

 I'm thinking about how much utility this actually provides compared to `docker volume rm clickhouse`. It might be better to just wipe the slate clean for an entire service, not a single docker container as that can be done pretty easily through native docker

IanWoodard: So a nuke command?

IanWoodard:
Closing in favor of just having a nuke command in the future

PR #34, opened 2024-09-18
feat(status): Add url

hubertdeng123:
Before:
Screenshot 2024-09-18 at 4 08 09 PM

After:
Screenshot 2024-09-18 at 4 08 02 PM

### DIFF
--- a/devservices/commands/status.py
+++ b/devservices/commands/status.py
@@ -47,9 +47,10 @@ def format_status_output(status_json: str) -> str:
 
         if ports:
             output.append("Ports:")
+            print(ports)
             for port in ports:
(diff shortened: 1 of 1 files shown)


devservices/commands/status.py
```diff
@@ -47,9 +47,10 @@ def format_status_output(status_json: str) -> str:
 
         if ports:
             output.append("Ports:")
+            print(ports)
```
IanWoodard, line 50: Is this intentional?

hubertdeng123: nope, good catch

IanWoodard (approved):

PR #36, opened 2024-09-20
feat(release): Build and upload binary

hubertdeng123:
This PR helps us publish a binary for devservices, aiding in a system wide installation of the cli tool. We're using PyOxidizer, which packages the python interpreter into the binary. 

1. builds a binary using PyOxidizer for linux and macos
2. uploads artifacts for the craft release job

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -44,6 +44,9 @@ jobs:
         with:
           python-version: 3.12
 
+      - name: Install pyoxidizer
+        run: pip install pyoxidizer==0.24.0
+
       - name: Install dev requirements
         run: pip install -r requirements-dev.txt
 
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -44,8 +44,8 @@ jobs:
         with:
           python-version: 3.12
 
-      - name: Install pyoxidizer
-        run: pip install pyoxidizer==0.24.0
+      - name: Install dev requirements
+        run: pip install -r requirements-dev.txt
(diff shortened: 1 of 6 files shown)


.github/workflows/build.yml
```diff
@@ -12,14 +14,60 @@ jobs:
...
       - uses: actions/setup-python@39cd14951b08e74b54015e9e001cdefcf80e669f # v5.1.1
         with:
           python-version: 3.12
+
+      - name: Install dev requirements
+        run: pip install -r requirements-dev.txt
+
       - name: "Prepare Artifacts"
-        run: |
-          pip install build
```
hubertdeng123, line 20: shouldn't install this generally, should pin to a version to prevent breakages

.github/workflows/build.yml
```diff
@@ -4,22 +4,107 @@ on:
...
+          path: binaries/devservices-linux
+          key: devservices-linux-${{ github.sha }}
+
+      - name: Restore macOS binary cache
+        uses: actions/cache@ab5e6d0c87105b4c9c2047343972218f562e4319 # v4.0.1
+        with:
+          path: binaries/devservices-darwin
+          key: devservices-darwin-${{ github.sha }}
+
+      - name: Upload combined artifacts
```
hubertdeng123, line 103: Uploading combined artifacts because with upload-artifact@v4, we cannot upload artifacts with the same name multiple times, which craft relies on

IanWoodard (approved):
LGTM

PR #41, opened 2024-09-25
feat(dep): Adding basic dependency management

IanWoodard:
Adding the first iteration of dependency management using git partialClone and sparse-checkout.
* Added basic dependency management logic (more improvements to come)
* Added basic testing for the dependency management logic including a basic fixture repo.


devservices/utils/dependencies.py
```diff
@@ -0,0 +1,132 @@
...
+            "--filter=blob:none",
+            "--no-checkout",
+            dependency.repo_link,
+            dependency_repo_dir,
+        ],
+        cwd=dependency_repo_dir,
+    )
+
+    # TODO: Should we include this with every command we run instead of setting config?
+    #       Doign so avoids the issue of updating the config if we inevitably change something.
```
hubertdeng123, line 103:
```suggestion
    #       Doing so avoids the issue of updating the config if we inevitably change something.
```

devservices/utils/dependencies.py
```diff
@@ -0,0 +1,132 @@
...
+        cwd=dependency_repo_dir,
+        stderr=subprocess.PIPE,
+    ).strip()
+
+    if local_commit == remote_commit:
+        # Already up-to-date, don't pull anything
+        return
+
+    # If it's not up-to-date, checkout the latest changes
+    _run_command(["git", "checkout", "FETCH_HEAD"], cwd=dependency_repo_dir)
```
hubertdeng123, line 80: If for some reason the state of the repo is messed up, things can go wrong here. I wonder if we can use some mechanism to make sure untracked files and such are cleaned up before doing this.

```
git clean -f
```

hubertdeng123 (approved):

devservices/utils/dependencies.py
IanWoodard: https://github.com/getsentry/devservices/issues/43

PR #49, opened 2024-09-30
chore(sentry): Add sentry environment support

IanWoodard:
Enable us to set a different sentry environment for development to avoid contaminating the production sentry environment data.


.gitignore
```diff
@@ -3,3 +3,4 @@
 __pycache__/
 
 .coverage
+.env
```
hubertdeng123, line 6: nit: newline

hubertdeng123 (approved):

PR #53, opened 2024-09-30
feat(docker-compose): Check version

hubertdeng123:


devservices/utils/docker_compose.py
```diff
@@ -1,14 +1,61 @@
...
+        )
+
+    if docker_compose_version is None:
+        print("Unable to detect docker compose version")
+        exit(1)
+    elif version.parse(docker_compose_version) < version.parse(
+        MINIMUM_DOCKER_COMPOSE_VERSION
+    ):
+        print("Docker compose version unsupported, please upgrade to >= 2.21.0")
+        exit(1)
```
IanWoodard, line 56: What are your thoughts on raising specific (and custom) exceptions rather than simply exiting and let the code that calls this figure out what it will do with the exception?

tests/utils/test_docker_compose.py
```diff
@@ -0,0 +1,48 @@
...
+@mock.patch("builtins.print")
+def test_check_docker_compose_version_unsupported(
+    mock_print: mock.Mock, mock_run: mock.Mock
+) -> None:
+    mock_run.return_value.stdout = "2.20.0-desktop.1\n"
+    with pytest.raises(SystemExit):
+        check_docker_compose_version()
+        mock_print.assert_called_with(
+            "Docker compose version unsupported, please upgrade to >= 2.21.0"
+        )
```
IanWoodard, line 28: Continuation of above: it would be nice to catch a specific exception rather than a system exit which I feel should happen in the code that calls this rather than the check function directly.

devservices/utils/docker_compose.py
```diff
@@ -1,14 +1,61 @@
...
+        else:
+            docker_compose_version = None
+
+    except subprocess.CalledProcessError as e:
+        raise DockerComposeError(
+            command=" ".join(cmd),
+            returncode=e.returncode,
+            stdout=e.stdout,
+            stderr=e.stderr,
+        )
```
IanWoodard, line 47: Could we limit what is in the try block to just what is necessary?

IanWoodard (approved):

PR #55, opened 2024-10-01
feat(list-services): Add status of services

hubertdeng123:
This changes the default behavior of this command to display running services. We also give people the option to display all services with -a/--all. 

with no --all option
Screenshot 2024-10-02 at 1 49 04 PM
Screenshot 2024-10-02 at 1 48 32 PM

with --all option
Screenshot 2024-10-02 at 1 50 34 PM


devservices/utils/docker_compose.py
```diff
@@ -9,6 +9,23 @@
 from devservices.utils.services import Service
 
 
+def get_active_docker_compose_projects() -> list[str]:
```
hubertdeng123, line 12: The main motivation behind this PR is this function. It will be used to help us with dependency management when bringing down services that have common dependencies.

devservices/commands/list_services.py
```diff
@@ -5,12 +5,19 @@
...
 
 from devservices.utils.devenv import get_coderoot
+from devservices.utils.docker_compose import get_active_docker_compose_projects
 from devservices.utils.services import get_local_services
 
 
 def add_parser(subparsers: _SubParsersAction[ArgumentParser]) -> None:
     parser = subparsers.add_parser(
-        "list-services", help="List the services installed locally"
+        "list-services", help="List the services installed locally", aliases=["ls"]
```
IanWoodard, line 14: I don't know how I feel as using `ls` as an alias here

IanWoodard (commented):
I feel like this functionality better fits in the status command, what are your thoughts? I feel that list-services should simply determine which services can be run and not worry about status, since we already have a command for that.

hubertdeng123:
Given the structure of our CLI, I think this is better. Generally, our cli follows this pattern:
1. `devservices command service_name`
2. `devservices command`

For 2, our cli tries to automatically detect a `devservices/config.yml` file in the current directory before proceeding. Adding the functionality in the PR to the status command would break this workflow, as I'm assuming that `devservices status` would give you the list of running services.

Moreover, the more I thought about it, I don't believe people will be using `list-services` very often to view all the services that are installed. Most people will care about which services are up and running.

tests/commands/test_list_services.py
```diff
@@ -0,0 +1,97 @@
...
+@mock.patch(
+    "devservices.utils.docker_compose.subprocess.run",
+    return_value=Namespace(stdout="\nexample-service\n"),
+)
+def test_list_running_services(
+    mock_run: mock.Mock, tmp_path: Path, capsys: pytest.CaptureFixture[str]
+) -> None:
+    with mock.patch(
+        "devservices.commands.list_services.get_coderoot",
+        return_value=str(tmp_path),
```
IanWoodard, line 23: nite: Could we put the coderoot in tmp_path / "code"?

devservices/commands/list_services.py
```diff
@@ -21,11 +28,30 @@ def list_services(args: Namespace) -> None:
...
 
     if not services:
         print("No services found")
         return
 
-    print("Services installed locally:")
-    for service in services:
-        print("-", service.name, f"({service.repo_path})")
+    services_to_show = (
+        services if args.all else [s for s in services if s.name in running_projects]
```
IanWoodard, line 38: Won't relying on docker compose project names to map to service names introduce some jagged edges? I can see this working fine in a lot of cases, but I can also see this biting up down the road. Any thoughts on this?

IanWoodard (approved):

PR #56, opened 2024-10-02
feat(dependency): Adding dependency management to commands

IanWoodard:
Adding the first layer of dependency management to commands. This includes pulling in one level of remote dependencies for services. What this means is that when running `start`, we first find and install any required remote dependencies for the given service. All other commands that rely on docker compose first check to see if the dependencies exist locally (as they should) since we want to re-use the same dependencies we installed/updated when we ran `start`. There are potential edge-cases where the service may have been started but the dependencies don't exist locally (not sure how this would happen yet, but it's likely possible). In this case, we have a basic fallback to simply re-install the dependencies, which while not ideal, since the dependencies could have been updated since running start, it is a good initial mitigation effort that we can further improve down the road.


devservices/utils/docker_compose.py
```diff
@@ -2,16 +2,38 @@
...
+    else:
+        are_dependencies_valid = verify_local_dependencies(dependencies)
+        if not are_dependencies_valid:
+            # TODO: Figure out how to handle this case as installing dependencies may not be the right thing to do
+            #       since the dependencies may have changed since the service was started.
+            install_dependencies(dependencies)
+    relative_local_dependency_directory = os.path.relpath(
+        DEVSERVICES_LOCAL_DEPENDENCIES_DIR, service.repo_path
+    )
+    with tempfile.NamedTemporaryFile(mode="w", delete=False) as temp_env_file:
```
hubertdeng123, line 32: Why do we have the delete=False in the context manager? We could also get rid of the finally block where we remove the temp env file if that is not set. It would be nice to have automatic cleanup throughout this PR

devservices/utils/dependencies.py
```diff
@@ -10,17 +10,37 @@
...
+    remote_configs = _get_remote_configs(dependencies)
+
+    # Short circuit to avoid doing unnecessary work
+    if len(remote_configs) == 0:
+        return True
+
+    if not os.path.exists(DEVSERVICES_LOCAL_DEPENDENCIES_DIR):
+        return False
+
+    return all(
```
hubertdeng123, line 29: Thinking of a case here where there is corruption in the local dependencies config, how do we detect that?

IanWoodard: I agree, but let's hold off on addressing it until a later PR

IanWoodard: https://github.com/getsentry/devservices/issues/58

devservices/utils/docker_compose.py
IanWoodard: Good call

IanWoodard: Good call, addressed!

hubertdeng123 (approved):

PR #64, opened 2024-10-03
feat(docker-compose): Simplify docker compose environment injection

### LINKED ISSUE #61: Avoid creating temp files for env variable injection
Investigate using env variable injection through subprocess instead of temp variables

IanWoodard:
Resolves #61


hubertdeng123 (approved):
🤩

PR #75, opened 2024-10-17
feat(commands): Add update command

hubertdeng123:
This adds the update command to devservices

For global installs, this will download the latest binary off of GitHub, detect the correct binary for linux/darwin operating systems, and automatically install. It will also automatically check for update after each invocation of devservices.

For installs in virtual environments, it prints instructions to update the requirements.txt.

### DIFF
--- a/devservices/main.py
+++ b/devservices/main.py
@@ -14,6 +14,8 @@ from devservices.commands import logs
 from devservices.commands import start
 from devservices.commands import status
 from devservices.commands import stop
+from devservices.commands import update
+from devservices.commands.check_for_update import check_for_update
 from devservices.utils.docker_compose import check_docker_compose_version
 
 sentry_environment = (
@@ -58,6 +60,7 @@ def main() -> None:
     list_services.add_parser(subparsers)
     status.add_parser(subparsers)
     logs.add_parser(subparsers)
+    update.add_parser(subparsers)
 
     args = parser.parse_args()
 
@@ -67,6 +70,14 @@ def main() -> None:
             args.func(args)
     else:
         parser.print_help()
+
+    if args.command != "update":
+        newest_version = check_for_update(metadata.version("devservices"))
+        if newest_version != metadata.version("devservices"):
+            print(
+                f"\n\033[93mWARNING: A new version of devservices is available: {newest_version}\033[0m"
+            )
+            print("To update, run: \033[1mdevservices update\033[0m")
 
 
 if __name__ == "__main__":
--- a/devservices/utils/docker_compose.py
+++ b/devservices/utils/docker_compose.py
@@ -3,12 +3,8 @@ from __future__ import annotations
 import os
 import platform
 import re
-import shutil
 import subprocess
-import tempfile
-import time
 from typing import cast
-from urllib.request import urlretrieve
 
 from packaging import version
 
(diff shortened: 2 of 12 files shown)


devservices/main.py
```diff
@@ -68,6 +71,14 @@ def main() -> None:
     else:
         parser.print_help()
 
+    if args.command != "update":
```
hubertdeng123, line 74: I can see a major downside to this:

If user is not connected to internet, this will error out and cause spam. I think it is ok to leave it as is and put a workaround in later though

hubertdeng123: Ultimately, it is far more useful to notify users to update if available

devservices/utils/docker_compose.py
```diff
@@ -60,62 +57,30 @@ def install_docker_compose() -> None:
...
+        binary_name_with_extension = f"docker-compose-darwin-{arch}"
     else:
         raise DockerComposeInstallationError(f"Unsupported operating system: {system}")
 
-    url = f"https://github.com/docker/compose/releases/download/v{MINIMUM_DOCKER_COMPOSE_VERSION}/{binary_name}"
-
-    # Create a temporary directory
-    with tempfile.TemporaryDirectory() as temp_dir:
-        temp_file = os.path.join(temp_dir, "docker-compose")
+    url = f"https://github.com/docker/compose/releases/download/v{MINIMUM_DOCKER_COMPOSE_VERSION}/{binary_name_with_extension}"
```
IanWoodard, line 70: Should we move the base url to a constant or a formatted string constant?

devservices/utils/install_binary.py
```diff
@@ -0,0 +1,50 @@
...
+import time
+from urllib.request import urlretrieve
+
+
+def install_binary(
+    binary_name: str,
+    exec_path: str,
+    version: str,
+    url: str,
+    exception_type: type[Exception],
```
IanWoodard, line 15: What are your thoughts on having a generic install binary error that we can catch and throw other, more specific errors from?

devservices/utils/install_binary.py
```diff
@@ -0,0 +1,50 @@
...
+                    )
+                    time.sleep(retry_delay_seconds)
+                else:
+                    raise exception_type(
+                        f"Failed to download {binary_name} after {max_retries} attempts: {e}"
+                    )
+
+        # Make the binary executable
+        try:
+            os.chmod(temp_file, 0o755)
```
IanWoodard, line 41: Thoughts on making this a constant?

devservices/main.py
```diff
@@ -68,6 +71,14 @@ def main() -> None:
...
+        newest_version = check_for_update(metadata.version("devservices"))
```
IanWoodard, line 75: Seems reasonable for now, but let's try to stop this from blocking. Maybe asyncio or something similar could be useful here.

devservices/utils/install_binary.py
```diff
@@ -0,0 +1,50 @@
...
+        except Exception as e:
```
IanWoodard, line 42: Can we catch more specific exceptions here?

devservices/utils/install_binary.py
```diff
@@ -0,0 +1,50 @@
...
+            raise exception_type(f"Failed to set executable permissions: {e}")
+
+        try:
+            shutil.move(temp_file, exec_path)
+        except Exception as e:
```
IanWoodard, line 47: Same as above

devservices/utils/install_binary.py
hubertdeng123: If we catch more specific exceptions here, we'll still need to catch general exceptions as well though. I can see us adding better print statements for our users when catching more specific exceptions which I could do if that is what you are saying

IanWoodard (approved):

PR #82, opened 2024-10-23
ref(dependencies): Providing information about installed remote dependencies after installing

IanWoodard:
To enable us to properly manage start and stop (as well as other commands) with nested dependencies, we need to know what nested service dependencies we need to start/stop separately and what dependencies may be shared between services.

### DIFF
--- a/devservices/utils/docker_compose.py
+++ b/devservices/utils/docker_compose.py
@@ -19,6 +19,7 @@ from devservices.constants import MINIMUM_DOCKER_COMPOSE_VERSION
 from devservices.exceptions import BinaryInstallError
 from devservices.exceptions import DockerComposeError
 from devservices.exceptions import DockerComposeInstallationError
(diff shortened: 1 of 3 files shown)


devservices/utils/docker_compose.py
```diff
@@ -151,13 +152,17 @@ def run_docker_compose_command(
...
     else:
         are_dependencies_valid = verify_local_dependencies(dependencies)
         if not are_dependencies_valid:
             # TODO: Figure out how to handle this case as installing dependencies may not be the right thing to do
             #       since the dependencies may have changed since the service was started.
-            install_dependencies(dependencies)
+            remote_dependencies = install_dependencies(dependencies)
+        else:
+            remote_dependencies = get_installed_remote_dependencies(dependencies)
+    # TODO: remove
```
hubertdeng123, line 164: should this be removed?

hubertdeng123 (approved):

PR #84, opened 2024-10-23
feat(purge): adding purge command

### LINKED ISSUE #71: Clean up local dependencies
If local dependencies are somehow corrupt, let's provide a command to clean those up so that people can start from scratch.

IanWoodard:
Fixes #71. Adds a new `purge` command that clears the user's local devservices cache.

### DIFF
--- a/devservices/commands/purge.py
+++ b/devservices/commands/purge.py
@@ -19,7 +19,7 @@ def purge(args: Namespace) -> None:
     if os.path.exists(DEVSERVICES_CACHE_DIR):
         try:
             shutil.rmtree(DEVSERVICES_CACHE_DIR)
-        except (FileNotFoundError, PermissionError) as e:
+        except PermissionError as e:
             print(f"Failed to purge cache: {e}")
             exit(1)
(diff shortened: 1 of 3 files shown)


devservices/commands/purge.py
```diff
@@ -0,0 +1,25 @@
...
+    parser = subparsers.add_parser("purge", help="Purge the local devservices cache")
+    parser.set_defaults(func=purge)
+
+
+def purge(args: Namespace) -> None:
+    """Purge the local devservices cache."""
+    if os.path.exists(DEVSERVICES_CACHE_DIR):
+        try:
+            shutil.rmtree(DEVSERVICES_CACHE_DIR)
+        except (FileNotFoundError, PermissionError) as e:
```
hubertdeng123, line 22: Do we really expect file not found error if the path exists? Anyhow, I think file not found error and permission error should be treated differently. If file is not found, it's not really an error in this case since the cache would then be "purged". If there is a permission error, the cache is not "purged" at all

hubertdeng123 (approved):

PR #86, opened 2024-10-24
feat(devservices): Run commands for nested dependencies by project

hubertdeng123:
This adds support for running docker commands for nested remote dependencies. Each docker compose command is run  with its own appropriate service name as its project.  

For example:
```
Grandparent (grandparent-service)
--> Parent (parent-service)
----> Child (child-service-1, child-service-2)
```
Docker Compose projects:
Grandparent: 
- grandparent-service

Parent:
- parent-service

Child:
- child-service-1
- child-service-2

### DIFF
--- a/devservices/utils/docker_compose.py
+++ b/devservices/utils/docker_compose.py
@@ -32,9 +32,7 @@ from devservices.utils.services import Service
 def get_active_docker_compose_projects() -> list[str]:
     cmd = ["docker", "compose", "ls", "-q"]
     try:
-        running_projects = subprocess.run(
-            cmd, check=True, capture_output=True, text=True
-        ).stdout
+        running_projects = subprocess.check_output(cmd, text=True)
     except subprocess.CalledProcessError as e:
         raise DockerComposeError(
             command=" ".join(cmd),
@@ -203,12 +201,15 @@ def _get_docker_compose_commands_to_run(
     )
     # Sort the remote dependencies by service name to ensure a deterministic order
     for dependency in sorted(remote_dependencies, key=lambda x: x.service_name):
+        # TODO: Consider passing in service config in InstalledRemoteDependency instead of loading it here
         dependency_service_config = load_service_config_from_file(dependency.repo_path)
         dependency_config_path = os.path.join(
             dependency.repo_path, DEVSERVICES_DIR_NAME, CONFIG_FILE_NAME
         )
-        services_defined = _get_non_remote_services(dependency_config_path, current_env)
-        services_to_use = services_defined.intersection(
+        non_remote_services = _get_non_remote_services(
+            dependency_config_path, current_env
+        )
+        services_to_use = non_remote_services.intersection(
             set(dependency_service_config.modes[dependency.mode])
         )
         docker_compose_commands.append(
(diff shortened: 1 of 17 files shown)


devservices/utils/docker_compose.py
```diff
@@ -148,20 +151,103 @@ def check_docker_compose_version() -> None:
...
+        "compose",
+        "-f",
+        service_config_path,
+        "config",
+        "--services",
+    ]
+    try:
+        config_services = subprocess.run(
+            config_command, capture_output=True, text=True, env=current_env
+        ).stdout
```
IanWoodard, line 169: Should we use `subprocess.check_output` instead?

devservices/utils/docker_compose.py
```diff
@@ -148,20 +151,103 @@ def check_docker_compose_version() -> None:
...
+    except subprocess.CalledProcessError as e:
+        raise DockerComposeError(
+            command=" ".join(config_command),
+            returncode=e.returncode,
+            stdout=e.stdout,
+            stderr=e.stderr,
+        ) from e
+    services_defined_in_config = set(config_services.splitlines())
+    return services_defined_in_config
```
IanWoodard, line 178: Nit: We could consolidate these lines

devservices/utils/docker_compose.py
```diff
@@ -148,20 +151,103 @@ def check_docker_compose_version() -> None:
...
+            "-f",
+            config_path,
+            command,
+        ]
+        + sorted(list(services_to_use))  # Sort the services to prevent flaky tests
+        + options
+    )
+    # Sort the remote dependencies by service name to ensure a deterministic order
+    for dependency in sorted(remote_dependencies, key=lambda x: x.service_name):
+        dependency_service_config = load_service_config_from_file(dependency.repo_path)
```
IanWoodard, line 206: Ideally we wouldn't have to re-look up the service config since we could pass it via the InstalledRemoteDependency. The main issue preventing that right now is that it is a set and the config isn't hashable.

devservices/utils/docker_compose.py
```diff
@@ -148,20 +151,103 @@ def check_docker_compose_version() -> None:
...
+        dependency_config_path = os.path.join(
+            dependency.repo_path, DEVSERVICES_DIR_NAME, CONFIG_FILE_NAME
+        )
+        services_defined = _get_non_remote_services(dependency_config_path, current_env)
```
IanWoodard, line 210: nit: Could we use similar naming as the function? Something like `non_remote_services`

devservices/utils/docker_compose.py
```diff
@@ -148,20 +151,103 @@ def check_docker_compose_version() -> None:
...
+        docker_compose_commands.append(
+            create_docker_compose_command(
+                dependency_service_config.service_name,
+                dependency_config_path,
+                services_to_use,
+            )
+        )
+
+    # Add docker compose command for the top level service
+    services_defined = _get_non_remote_services(service_config_file_path, current_env)
```
IanWoodard, line 223: Same comment as above

devservices/utils/docker_compose.py
hubertdeng123: Something we should talk about later, will leave a TODO

devservices/utils/docker_compose.py
hubertdeng123: guess that would be a bit nicer, done

IanWoodard (approved):

PR #99, opened 2024-10-31
chore(errors): Adding graceful handling when docker daemon isn't running

### LINKED ISSUE #95: Handle docker not running more gracefully
Currently, when trying to run most commands in devservices without docker running, it will fail in a very dramatic way. Ideally we should more concisely relay the cause to the user to promote better UX.

IanWoodard:
Fixes #95


devservices/utils/docker_compose.py
```diff
@@ -104,11 +105,23 @@ def install_docker_compose() -> None:
 
 
 def check_docker_compose_version() -> None:
-    cmd = ["docker", "compose", "version", "--short"]
+    # Make sure the Docker daemon is running
```
hubertdeng123, line 108: Would this be better to run in a standalone function `check_docker_daemon_running`?

hubertdeng123 (approved):

PR #101, opened 2024-10-31
chore(errors): Improving the error msgs for dependency errors

### LINKED ISSUE #96: Add better error message for dependency errors
Currently, we just drop the entire stack trace in the terminal rather than giving them a more meaningful error message such as which dependencies are having issues.

IanWoodard:
Fixes #96


hubertdeng123 (approved):
LGTM! Do any tests here need to be updated to be more thorough?

PR #103, opened 2024-10-31
feat(state): Add state with sqlite db

hubertdeng123:
This adds functionality to add a local file that stores state across devservices invocations. It uses sqlite to store and handle started_services


devservices/constants.py
```diff
@@ -11,6 +11,7 @@
 DEVSERVICES_LOCAL_DIR = os.path.expanduser("~/.local/share/sentry-devservices")
 DEVSERVICES_DEPENDENCIES_CACHE_DIR = os.path.join(DEVSERVICES_CACHE_DIR, "dependencies")
 DEVSERVICES_DEPENDENCIES_CACHE_DIR_KEY = "DEVSERVICES_DEPENDENCIES_CACHE_DIR"
+DB_FILE = os.path.join(DEVSERVICES_LOCAL_DIR, "state")
```
IanWoodard, line 14: What do you think about STATE_DB_FILE?

IanWoodard (approved):

PR #104, opened 2024-11-01
feat(devservices): Proper stop logic to account for shared remote services

hubertdeng123:
If remote services are shared between running services, `devservices stop` will not stop those services. Also, users will not be able to stop services that are not already running.


devservices/commands/stop.py
```diff
@@ -33,6 +33,11 @@ def stop(args: Namespace) -> None:
     # TODO: allow custom modes to be used
     mode_to_stop = "default"
     mode_dependencies = modes[mode_to_stop]
+    state = State()
+    started_services = state.get_started_services()
+    if service.name not in started_services:
```
IanWoodard, line 38: Not relevant to this PR, but we should figure out how to better handle edge cases and corrupt state here. User's could use the nuke command to clear their state and running services if their state gets messed up.

IanWoodard (approved):

PR #108, opened 2024-11-01
fix(list-services): Use state for finding status of list services

hubertdeng123:
For `list-services`, we are incorrectly marking local services as running when services are brought up with remote dependencies. 

Say for example snuba is brought up that depends on shared-redis and shared-kafka whose definitions are pulled in remotely. `devservices list-services` would show local copies of shared-redis and shared-kafka as running. This fixes that

### DIFF
--- a/devservices/utils/docker_compose.py
+++ b/devservices/utils/docker_compose.py
@@ -29,21 +29,6 @@ from devservices.utils.install_binary import install_binary
 from devservices.utils.services import Service
 
 
-def get_active_docker_compose_projects() -> list[str]:
-    cmd = ["docker", "compose", "ls", "-q"]
-    try:
-        running_projects = subprocess.check_output(cmd, text=True)
-    except subprocess.CalledProcessError as e:
-        raise DockerComposeError(
-            command=" ".join(cmd),
-            returncode=e.returncode,
-            stdout=e.stdout,
-            stderr=e.stderr,
-        )
-    # docker compose ls always returns newline delimited string with an extra newline at the end
-    return running_projects.split("\n")[:-1]
-
-
 def install_docker_compose() -> None:
     # Determine the platform
     system = platform.system()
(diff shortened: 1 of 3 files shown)


devservices/utils/docker_compose.py
```diff
@@ -29,21 +29,6 @@
 from devservices.utils.services import Service
 
 
-def get_active_docker_compose_projects() -> list[str]:
```
hubertdeng123, line 32: 👋

tests/commands/test_list_services.py
```diff
@@ -7,20 +7,19 @@
...
-    mock_run: mock.Mock, tmp_path: Path, capsys: pytest.CaptureFixture[str]
+    tmp_path: Path, capsys: pytest.CaptureFixture[str]
 ) -> None:
     with mock.patch(
         "devservices.commands.list_services.get_coderoot",
         return_value=str(tmp_path / "code"),
-    ):
+    ), mock.patch("devservices.utils.state.STATE_DB_FILE", str(tmp_path / "state")):
+        state = State()
+        state.add_started_service("example-service", "default")
```
IanWoodard, line 22: Feel free to punt this for now, but ideally we should be calling start (not necessarily actually starting the service, but doing whatever state and other misc things that are required)

tests/commands/test_list_services.py
```diff
@@ -52,17 +51,13 @@ def test_list_running_services(
...
-    mock_run: mock.Mock, tmp_path: Path, capsys: pytest.CaptureFixture[str]
-) -> None:
+def test_list_all_services(tmp_path: Path, capsys: pytest.CaptureFixture[str]) -> None:
     with mock.patch(
         "devservices.commands.list_services.get_coderoot",
         return_value=str(tmp_path / "code"),
-    ):
+    ), mock.patch("devservices.utils.state.STATE_DB_FILE", str(tmp_path / "state")):
+        state = State()
+        state.add_started_service("example-service", "default")
```
IanWoodard, line 60: Same comment as above

IanWoodard (approved):

tests/commands/test_list_services.py
hubertdeng123: seems out of scope for now but can consider in the future

Issue #123, opened 2024-11-08
Add debug mode


hubertdeng123:
This mode will provide more verbose output when running devservices commands.

PR #124, opened 2024-11-08
feat(logging): Add debug mode

hubertdeng123:
closes https://github.com/getsentry/devservices/issues/123


README.md
```diff
@@ -33,3 +33,11 @@ NOTE: service-name is an optional parameter. If not provided, devservices will a
...
 - `devservices update` Update devservices to the latest version.
 - `devservices purge`: Purge the local devservices cache.
+
+## Debug Mode
+
+To enable debug mode, set the `DEBUG` environment variable to `true`:
+
+```
+export DEBUG=true
+```
```
IanWoodard, line 43: What are your thoughts on also providing an optional debug flag for certain commands. The use case I see is running a specific command, it failing, and wanting to do a one-off debug for that command without wanting to go through and adding an entire env var.

hubertdeng123: I initially explored this path, but realized that it required refactoring quite a bit of code to support if we wanted to pass down the debug arg downstream. I looked at the diff and made a judgement call that it wasn't worth the changes it required.

PR #129, opened 2024-11-11
ref(docker-compose): Moving dependency logic out of docker compose

IanWoodard:
Starting to reduce the complexity of the run_docker_compose_command by decoupling dependency installation logic.

### DIFF
--- a/devservices/utils/docker_compose.py
+++ b/devservices/utils/docker_compose.py
@@ -224,8 +224,8 @@ def run_docker_compose_command(
     service: Service,
     command: str,
     mode_dependencies: list[str],
+    remote_dependencies: set[InstalledRemoteDependency],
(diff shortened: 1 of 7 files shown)


devservices/utils/docker_compose.py
```diff
@@ -229,24 +225,8 @@ def run_docker_compose_command(
     command: str,
     mode_dependencies: list[str],
     options: list[str] = [],
-    force_update_dependencies: bool = False,
+    remote_dependencies: set[InstalledRemoteDependency] = set(),
```
hubertdeng123, line 228: Seems like we should always be passing `remote_dependencies` here, so it shouldn't have a default

hubertdeng123 (approved):

PR #148, opened 2024-11-16
feat(start): Add more context to what is going on underneath the hood for start

hubertdeng123:
This will more clearly state the steps that are happening underneath the hood for starting a service.

Screenshot 2024-11-15 at 4 16 57 PM


devservices/constants.py
```diff
@@ -12,6 +12,7 @@
 DEVSERVICES_DEPENDENCIES_CACHE_DIR = os.path.join(DEVSERVICES_CACHE_DIR, "dependencies")
 DEVSERVICES_DEPENDENCIES_CACHE_DIR_KEY = "DEVSERVICES_DEPENDENCIES_CACHE_DIR"
 STATE_DB_FILE = os.path.join(DEVSERVICES_LOCAL_DIR, "state")
+DOCKER_COMPOSE_COMMAND_LENGTH = 7
```
IanWoodard, line 15: Is this the same across all docker compose commands or is it only true for up?

devservices/commands/up.py
```diff
@@ -79,10 +82,20 @@ def up(args: Namespace) -> None:
     state.add_started_service(service.name, mode)
 
 
+def _bring_up_dependency(
+    cmd: list[str], current_env: dict[str, str], status: Status, len_options: int
+) -> subprocess.CompletedProcess[str]:
+    # TODO: Get rid of these constants, we need a smarter way to determine the containers being brought up
+    for dependency in cmd[DOCKER_COMPOSE_COMMAND_LENGTH:-len_options]:
```
IanWoodard, line 89: I somewhat prefer the magic numbers if we can factor them out soon, but this is fine too

IanWoodard (approved):

devservices/constants.py
hubertdeng123: True for all

PR #151, opened 2024-11-18
ref(dependencies): Updating dependency management to handle modes

IanWoodard:
Currently, our dependency logic installs and returns all dependencies rather than just those needed for a specific mode. This PR changes that to be more efficient and accurate.

### DIFF
--- a/devservices/exceptions.py
+++ b/devservices/exceptions.py
@@ -67,6 +67,16 @@ class DockerComposeError(Exception):
         self.stderr = stderr
 
 
+class ModeDoesNotExistError(Exception):
+    """Raised when a mode does not exist."""
+
+    def __init__(self, mode: str):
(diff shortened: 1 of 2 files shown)


devservices/exceptions.py
```diff
@@ -67,6 +67,16 @@ def __init__(self, command: str, returncode: int, stdout: str, stderr: str):
         self.stderr = stderr
 
 
+class ModeDoesNotExistError(Exception):
+    """Raised when a mode does not exist."""
+
+    def __init__(self, mode: str):
```
hubertdeng123, line 73: Should we also pass in service name?

IanWoodard: Done

hubertdeng123 (approved):

PR #160, opened 2024-11-20
fix(up): Restart services when switching modes

IanWoodard:
Currently, there is a bug where if you bring a service up in a certain mode, then bring it up in a different mode, there not only could be extraneous dependencies left running, but also the internal state does not update meaning bringing the service back down will likely not work correctly. As a short term fix for this, we opted to check the current running mode and the desired mode and if they differ, but for the same service, we restart the service and bring it up in the desired mode. Longer term we plan on making the logic more intelligent and able to only bring up and down what is necessary to make it faster and more efficient.


devservices/commands/up.py
```diff
@@ -63,8 +65,59 @@ def up(args: Namespace) -> None:
...
+        with Status() as status:
+            try:
+                remote_dependencies_to_bring_down = install_and_verify_dependencies(
+                    service, mode=running_mode
+                )
+            except DependencyError as de:
+                capture_exception(de)
+                status.failure(str(de))
+                exit(1)
+            except ModeDoesNotExistError as mde:
```
hubertdeng123, line 86: I don't think we should catch `ModeDoesNotExistError` here, since if a service is running in a certain mode we can assume it exists.

devservices/commands/up.py
```diff
@@ -63,8 +65,59 @@ def up(args: Namespace) -> None:
...
+                service=service,
+                remote_dependencies=remote_dependencies_to_bring_down,
+                current_env=current_env,
+                command="down",
+                options=[],
+                service_config_file_path=service_config_file_path,
+                mode_dependencies=running_mode_dependencies,
+            )
+
+            cmd_outputs = []
```
hubertdeng123, line 107: Looks like we don't need to store cmd_outputs, so can this be removed?

devservices/commands/up.py
```diff
@@ -63,8 +65,59 @@ def up(args: Namespace) -> None:
...
+    running_mode = state.get_mode_for_service(service.name) or "default"
+
+    # TODO: Remove this once we properly handle mode switching
+    if service.name in started_services and running_mode != mode:
+        console.warning(
+            f"Service '{service.name}' is already running in mode: '{running_mode}', restarting in mode: '{mode}'"
+        )
+        with Status() as status:
+            try:
+                remote_dependencies_to_bring_down = install_and_verify_dependencies(
```
hubertdeng123, line 79: nit: `remote_dependencies` since technically this could include remote dependencies we don't want to bring down

hubertdeng123 (approved):

devservices/commands/up.py
IanWoodard: I think we should catch it, but in this case capture it as well to be alerting if it happens

PR #163, opened 2024-11-21
feat(dependency): Adding dependency graph to start services

IanWoodard:
Currently, our logic does not handle bringing up services in the correct order such that dependencies are started before their dependents. This change addresses that by constructing a dependency graph and using reverse topological ordering to determine the ordering in which we should bring services up.

### DIFF
--- a/devservices/utils/dependencies.py
+++ b/devservices/utils/dependencies.py
@@ -14,6 +14,7 @@ from typing import TypeGuard
 from devservices.configs.service_config import Dependency
 from devservices.configs.service_config import load_service_config_from_file
 from devservices.configs.service_config import RemoteConfig
+from devservices.configs.service_config import ServiceConfig
 from devservices.constants import CONFIG_FILE_NAME
 from devservices.constants import DEPENDENCY_CONFIG_VERSION
 from devservices.constants import DEPENDENCY_GIT_PARTIAL_CLONE_CONFIG_OPTIONS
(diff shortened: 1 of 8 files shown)


tests/utils/test_dependencies.py
```diff
@@ -1580,3 +1581,384 @@ def test_install_and_verify_dependencies_mode_does_not_exist(tmp_path: Path) ->
...
+                    "description": "parent-service",
+                    "remote": {
+                        "repo_name": "parent-service",
+                        "repo_link": f"file://{parent_service_repo_path}",
+                        "branch": "main",
+                    },
+                },
+                "grandparent-service": {
+                    "description": "grandparent-service",
+                },
```
IanWoodard, line 1894: I see why we have this, but aren't we planning on moving away from this pattern and toward having different naming for self-referenced containerized versions of the service?

devservices/utils/dependencies.py
```diff
@@ -35,6 +36,58 @@
...
+            for dependency in self.graph[service_name]:
+                in_degree[dependency] += 1
+
+        queue = [
+            service_name for service_name in self.graph if in_degree[service_name] == 0
+        ]
+        topological_order = list()
+
+        while queue:
+            service_name = queue.pop(0)
```
hubertdeng123, line 73: I think this is O(n) actually, and perhaps we should use deque() instead and use `popleft`

IanWoodard: Good idea!

hubertdeng123 (approved):

PR #165, opened 2024-11-26
feat(tests): Test status command

hubertdeng123:
This adds tests for status command

### DIFF
--- a/tests/commands/test_status.py
+++ b/tests/commands/test_status.py
@@ -30,6 +30,11 @@ def test_status_service_not_found(
         status(args)
 
     assert exc_info.value.code == 1
+
+    mock_find_matching_service.assert_called_once_with("nonexistent-service")
+    mock_install_and_verify_dependencies.assert_not_called()
+    mock_status.assert_not_called()
+
     captured = capsys.readouterr()
     assert "Service not found" in captured.out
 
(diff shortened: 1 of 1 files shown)


tests/commands/test_status.py
```diff
@@ -0,0 +1,153 @@
...
+    captured = capsys.readouterr()
+    assert (
+        """Service: test-service
+
+----------------------------------------
+test-service
+Container: test-container
+Status: running
+Health: healthy
+Uptime: 2 days ago
```
IanWoodard, line 146: This is unrelated and we can/should address this separately, but what are your thoughts on referring to uptime as "2 days" rather than "2 days ago"? I can see it going both ways, but I think the former is slightly cleaner.

tests/commands/test_status.py
```diff
@@ -0,0 +1,153 @@
...
+    mock_install_and_verify_dependencies.side_effect = DependencyError(
+        "Dependency error", repo_link=str(tmp_path), branch="main"
+    )
+
+    with pytest.raises(SystemExit) as exc_info:
+        status(args)
+
+    assert exc_info.value.code == 1
+    captured = capsys.readouterr()
+    assert "Dependency error" in captured.out
```
IanWoodard, line 68:
```suggestion
    mock_install_and_verify_dependencies.side_effect = DependencyError(
        repo_name="test-service", repo_link=str(tmp_path), branch="main"
    )

    with pytest.raises(SystemExit) as exc_info:
        status(args)

    assert exc_info.value.code == 1
    captured = capsys.readouterr()
    assert f"DependencyError: test-service ({str(tmp_path)}) on main" in captured.out
```

tests/commands/test_status.py
hubertdeng123: "2 days ago" is actually the text word for word what we get from docker/docker compose, so I think we should keep it like this

IanWoodard (approved):

PR #166, opened 2024-11-27
feat(services): Adding test for skipping non-devservices repos

IanWoodard:
Adding an additional test to cover the case where we skip over local repos when they don't contain a devservices config.

### DIFF
--- a/testing/resources/non-devservices-repo/README.md
+++ b/testing/resources/non-devservices-repo/README.md
@@ -1,1 +1,1 @@
-# This is a repo with a devservices config
+# This is a repo without a devservices config
--- a/testing/resources/non-devservices-repo/README.md
+++ b/testing/resources/non-devservices-repo/README.md
@@ -1,0 +1,1 @@
+# This is a repo with a devservices config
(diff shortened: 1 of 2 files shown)


testing/resources/non-devservices-repo/README.md
```diff
@@ -0,0 +1 @@
+# This is a repo with a devservices config
```
hubertdeng123, line 1: Does it have a devservices config though?🤔

tests/utils/test_services.py
```diff
@@ -57,3 +57,27 @@ def test_get_local_services_with_valid_config(tmp_path: Path) -> None:
...
+            "devservices.utils.dependencies.DEVSERVICES_DEPENDENCIES_CACHE_DIR",
+            str(tmp_path / "dependency-dir"),
+        ),
+        mock.patch(
+            "devservices.utils.services.get_coderoot",
+            return_value=str(mock_code_root),
+        ),
+    ):
+        os.makedirs(mock_code_root)
+        mock_devservices_repo_path = mock_code_root / "basic"
```
hubertdeng123, line 75: Should this be basic_repo given the repo name is basic_repo?

testing/resources/non-devservices-repo/README.md
IanWoodard: ope, good catch

tests/utils/test_services.py
IanWoodard: fair enough, will update

hubertdeng123 (approved):

PR #169, opened 2024-12-02
fix(sentry): Do not report to sentry if service is not found

hubertdeng123:
This adds additional noise, so let's not report to sentry if ServiceNotFound error is raised

### DIFF
--- a/devservices/commands/list_dependencies.py
+++ b/devservices/commands/list_dependencies.py
@@ -32,8 +32,11 @@ def list_dependencies(args: Namespace) -> None:
 
     try:
         service = find_matching_service(service_name)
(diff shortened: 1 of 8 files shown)


devservices/commands/list_dependencies.py
```diff
@@ -32,10 +32,13 @@ def list_dependencies(args: Namespace) -> None:
 
     try:
         service = find_matching_service(service_name)
-    except (ConfigError, ServiceNotFoundError) as e:
+    except ConfigError as e:
```
hubertdeng123, line 35: I also do not think we want to avoid reporting ConfigNotFoundError too, thoughts?

IanWoodard (approved):
Could you add some quick tests for this?

PR #170, opened 2024-12-04
feat(list-dependencies): list-dependencies shows mode of service that is running

hubertdeng123:
This change adds mode service is running in to the ls command


devservices/commands/list_services.py
```diff
@@ -47,7 +47,9 @@ def list_services(args: Namespace) -> None:
 
     for service in services_to_show:
         status = "running" if service.name in running_services else "stopped"
+        mode = state.get_mode_for_service(service.name)
```
IanWoodard, line 50: Should we handle potential errors for this?

IanWoodard (approved):

PR #172, opened 2024-12-05
fix(dependencies): Fixing dependency graph construction for simple modes

IanWoodard:
Fixing the dependency graph creation when a mode doesn't contain all of the dependencies and not all dependencies are downloaded. This bug happened because when we made the dependency graph we would traverse all dependencies, even the ones not included in the specified mode, meaning we would potentially traverse some that aren't installed leading to errors. This change addresses that issue by only traversing relevant dependencies for the given mode.

### DIFF
--- a/devservices/utils/dependencies.py
+++ b/devservices/utils/dependencies.py
@@ -531,14 +531,14 @@ def construct_dependency_graph(
 ) -> DependencyGraph:
     dependency_graph = DependencyGraph()
 
-    if modes is None:
-        modes = ["default"]
-
-    service_mode_dependencies = set()
-    for mode in modes:
-        service_mode_dependencies.update(service.config.modes.get(mode, []))
-
-    def _construct_dependency_graph(service_config: ServiceConfig) -> None:
+    def _construct_dependency_graph(
+        service_config: ServiceConfig, modes: list[str]
+    ) -> None:
+        if modes is None:
+            modes = ["default"]
+        service_mode_dependencies = set()
+        for mode in modes:
+            service_mode_dependencies.update(service_config.modes.get(mode, []))
         for dependency_name, dependency in service_config.dependencies.items():
             # Skip the dependency if it's not in the modes (since it may not be installed and we don't care about it)
             if (
@@ -549,7 +549,7 @@ def construct_dependency_graph(
             dependency_graph.add_edge(service_config.service_name, dependency_name)
             if _has_remote_config(dependency.remote):
                 dependency_config = get_remote_dependency_config(dependency.remote)
-                _construct_dependency_graph(dependency_config)
+                _construct_dependency_graph(dependency_config, [dependency.remote.mode])
 
-    _construct_dependency_graph(service.config)
+    _construct_dependency_graph(service.config, modes or ["default"])
     return dependency_graph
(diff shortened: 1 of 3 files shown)


devservices/utils/dependencies.py
```diff
@@ -526,15 +526,30 @@ def get_remote_dependency_config(remote_config: RemoteConfig) -> ServiceConfig:
...
+            ):
+                continue
             dependency_graph.add_edge(service_config.service_name, dependency_name)
             if _has_remote_config(dependency.remote):
                 dependency_config = get_remote_dependency_config(dependency.remote)
-                _construct_dependency_graph(dependency_config)
+                _construct_dependency_graph(dependency_config, [dependency.remote.mode])
 
-    _construct_dependency_graph(service.config)
+    _construct_dependency_graph(service.config, modes or ["default"])
```
hubertdeng123, line 554:
```suggestion
    _construct_dependency_graph(service.config, modes)
```

You have the check above if modes is None to use default, don't think it's needed

devservices/utils/dependencies.py
```diff
@@ -526,15 +526,30 @@ def get_remote_dependency_config(remote_config: RemoteConfig) -> ServiceConfig:
     return load_service_config_from_file(dependency_repo_dir)
 
 
-def construct_dependency_graph(service: Service) -> DependencyGraph:
+def construct_dependency_graph(
```
hubertdeng123, line 529: Should we add a test for this? Seems like it is possible by checking to ensure that dependencies not in modes should not appear in the graph

devservices/utils/dependencies.py
```diff
@@ -526,15 +526,30 @@ def get_remote_dependency_config(remote_config: RemoteConfig) -> ServiceConfig:
...
+    def _construct_dependency_graph(
+        service_config: ServiceConfig, modes: list[str]
+    ) -> None:
+        if modes is None:
+            modes = ["default"]
+        service_mode_dependencies = set()
+        for mode in modes:
+            service_mode_dependencies.update(service_config.modes.get(mode, []))
         for dependency_name, dependency in service_config.dependencies.items():
+            # Skip the dependency if it's not in the modes (since it may not be installed and we don't care about it)
```
hubertdeng123, line 543: This check will only happen at the top level service config right? Can we add a note to indicate that if true?

devservices/utils/dependencies.py
IanWoodard: Yep, good call

devservices/utils/dependencies.py
hubertdeng123: Wait, will there be an edge case where we have the scenario where a nested dependency is running in a mode where a remote dependency is not properly being skipped? Assume all dependencies here are remote dependencies

Service A:
Mode: migrations
Dependencies: [{name: "Service B", mode: "migrations"}, {name: "Service C", mode: "default"}]
modes: [{"default": ["Service B", "Service C"], "migrations": ["Service B"]}

Service B
Mode: migrations
Dependencies: [{name: "Service D", mode: "default"}, {"name: "Service E", mode: "default"}]
modes: [{"default": ["Service D", "Service E"], "migrations": ["Service D"]}

IanWoodard: That was left over code, I will fix it. It shouldn't only happen at the top level

IanWoodard: Fixed

PR #173, opened 2024-12-06
ref(modes): Refactoring modes to support multiple concurrent modes

IanWoodard:
Currently, we only support bringing up a service in a single mode at a time. After discussing it further with the team, we concluded that it would make more sense to allow users to run multiple modes simultaneously and to not have bringing up a sub-mode resulting in dependencies being brought down.

### DIFF
--- a/devservices/utils/state.py
+++ b/devservices/utils/state.py
@@ -35,17 +35,26 @@ class State:
         )
         self.conn.commit()
 
-    def add_started_service(self, service_name: str, mode: str) -> None:
+    def update_started_service(self, service_name: str, mode: str) -> None:
         cursor = self.conn.cursor()
         started_services = self.get_started_services()
-        if service_name in started_services:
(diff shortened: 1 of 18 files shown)


devservices/utils/state.py
```diff
@@ -35,17 +35,26 @@ def initialize_database(self) -> None:
...
+    def update_started_service(self, service_name: str, mode: str) -> None:
         cursor = self.conn.cursor()
         started_services = self.get_started_services()
-        if service_name in started_services:
+        active_modes = self.get_active_modes_for_service(service_name)
+        if service_name in started_services and mode in active_modes:
             return
-        cursor.execute(
-            """
-            INSERT INTO started_services (service_name, mode) VALUES (?, ?)
```
IanWoodard, line 45: We plan on renaming mode to modes once we figure out versioning. We also plan on making a more optimized table setup at that time as well.

hubertdeng123 (approved):

PR #175, opened 2024-12-06
fix(dependencies): Only install dependenices needed for the mode

IanWoodard:
This change updates our installation logic to only install dependencies necessary for a given mode even for nested dependencies which should help install be more efficient.

### DIFF
--- a/devservices/utils/dependencies.py
+++ b/devservices/utils/dependencies.py
@@ -355,7 +355,17 @@ def install_dependency(dependency: RemoteConfig) -> set[InstalledRemoteDependenc
                 branch=dependency.branch,
             ) from e
 
-    nested_dependencies = list(installed_config.dependencies.values())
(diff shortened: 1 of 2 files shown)


devservices/utils/dependencies.py
```diff
@@ -355,7 +355,17 @@ def install_dependency(dependency: RemoteConfig) -> set[InstalledRemoteDependenc
...
             ) from e
 
-    nested_dependencies = list(installed_config.dependencies.values())
+    if dependency.mode not in installed_config.modes:
+        raise ModeDoesNotExistError(
+            service_name=installed_config.service_name,
+            mode=dependency.mode,
+        )
+
+    nested_dependencies = [
```
hubertdeng123, line 364:
```suggestion
    active_nested_dependencies = [
```

hubertdeng123 (approved):

PR #178, opened 2024-12-10
feat(devservices): Add healthcheck wait condition and parallelize starting of containers

hubertdeng123:
This starts containers in parallel and makes sure devservices waits until containers are healthy before succeeding.


hubertdeng123:
I think relay sometimes doesn't pass the healthcheck, I may need to increase the retries there

joshuarli (commented):
imo `--wait` should be default

joshuarli (commented):
can we also define a redis healthcheck in sentry's devservices config? otherwise we could run into `Redis is loading the dataset in memory` during migrations in sentry sync.py

hubertdeng123:
> imo --wait should be default
I agree

> can we also define a redis healthcheck in sentry's devservices config? otherwise we could run into Redis is loading the dataset in memory during migrations in sentry sync.py

It is already defined:

https://github.com/getsentry/sentry-shared-redis/blob/9c15e58b7b254e2fd4621fb47d790c22d1eeaf3e/devservices/config.yml#L14

hubertdeng123:
Decided to use in-house logic for healtchecks because it seems like `--wait` is not reliable for relay unfortunately. Healthcheck is failing when the container itself clearly is displaying as healthy.  Spent a few hours trying to figure out the problem with docker compose, and decided this was better as in the future we can add logging to indicate specific services have passed these healthchecks.

devservices/utils/docker.py
```diff
@@ -30,13 +87,14 @@ def get_matching_containers(label: str) -> list[str]:
                 [
                     "docker",
                     "ps",
-                    "-q",
+                    "--format",
```
hubertdeng123, line 90: Docker container names are unique so this will be safe. Decided to make this change so that we could print a better status message if containers fail healthchecks

devservices/commands/up.py
```diff
@@ -144,8 +146,18 @@ def _up(
...
+    cmd_outputs = []
+    with concurrent.futures.ThreadPoolExecutor() as dependency_executor:
+        futures = [
+            dependency_executor.submit(
+                _bring_up_dependency, cmd, current_env, status, len(options)
+            )
+            for cmd in docker_compose_commands
+        ]
+        for future in concurrent.futures.as_completed(futures):
+            cmd_outputs.append(future.result())
```
joshuarli, line 158: you aren't doing anything with the results though? can just `_ = future.result()`

joshuarli, line 158: nit: since you're fire and forgetting just `_ = future.result()`

devservices/utils/docker.py
```diff
@@ -19,9 +25,60 @@ def check_docker_daemon_running() -> None:
...
+            ).strip()
+        except subprocess.CalledProcessError as e:
+            raise DockerError(
+                command=f"docker inspect -f '{{.State.Health.Status}}' {container_name}",
+                returncode=e.returncode,
+                stdout=e.stdout,
+                stderr=e.stderr,
+            ) from e
+
+        if result == "healthy" or result == "unknown":
```
joshuarli, line 68: for `unknown` (healthcheck doesn't exist?) we should probably emit a warning just to let people know there isn't a healthcheck available

IanWoodard, line 68:
```suggestion
        if result in ["healthy", "unknown"]:
```

hubertdeng123: That's fair, can definitely add this

devservices/utils/docker.py
```diff
@@ -19,9 +24,64 @@ def check_docker_daemon_running() -> None:
...
+                    "-f",
+                    "{{if .State.Health}}{{.State.Health.Status}}{{else}}unknown{{end}}",
+                    container_name,
+                ],
+                stderr=subprocess.DEVNULL,
+                text=True,
+            ).strip()
+        except subprocess.CalledProcessError as e:
+            raise DockerError(
+                command=f"docker inspect -f '{{.State.Health.Status}}' {container_name}",
```
IanWoodard, line 60: Should this match the command above completely?

devservices/utils/docker.py
```diff
@@ -19,9 +24,64 @@ def check_docker_daemon_running() -> None:
...
+                f"WARNING: Container {container_name} does not have a healthcheck"
+            )
+            return
+
+        # If not healthy, wait and try again
+        time.sleep(HEALTHCHECK_INTERVAL)
+
+    status.failure(
+        f"Container {container_name} did not become healthy within {HEALTHCHECK_TIMEOUT} seconds."
+    )
```
IanWoodard, line 79: Instead of using status here, could we throw an error that we can catch higher up and write a failure status up there?

devservices/exceptions.py
```diff
@@ -127,3 +127,16 @@ class FailedToSetGitConfigError(GitConfigError):
...
+    """Raised when a container is not healthy."""
+
+    def __init__(self, container_name: str, timeout: int):
+        self.container_name = container_name
+        self.timeout = timeout
+
+    def __str__(self) -> str:
+        return f"Container {self.container_name} did not become healthy within {self.timeout} seconds."
+
+    pass
```
IanWoodard, line 142: Do we need this?

devservices/utils/docker.py
```diff
@@ -19,9 +25,62 @@ def check_docker_daemon_running() -> None:
...
+                stderr=e.stderr,
+            ) from e
+
+        if result == "healthy":
+            return
+        elif result == "unknown":
+            status.warning(
+                f"WARNING: Container {container_name} does not have a healthcheck"
+            )
+            return
```
IanWoodard, line 73: Not saying we should go down this path, but an alternative would be to listen to docker events rather than polling. It could be worth looking into at some point.

tests/utils/test_docker.py
```diff
@@ -222,3 +225,223 @@ def test_stop_containers_remove_error(
...
+def test_check_all_containers_healthy_failure(
+    mock_check_output: mock.Mock,
+) -> None:
+    mock_status = mock.Mock()
+    mock_check_output.side_effect = ["healthy", "unhealthy", "unhealthy"]
+    with mock.patch("devservices.utils.docker.HEALTHCHECK_TIMEOUT", 2), mock.patch(
+        "devservices.utils.docker.HEALTHCHECK_INTERVAL", 1
+    ):
+        with pytest.raises(ContainerHealthcheckFailedError):
+            check_all_containers_healthy(mock_status, ["container1", "container2"])
```
IanWoodard, line 410: We should use freezegun for these tests to avoid having time.sleep impact the duration of the test.

tests/commands/test_up.py
```diff
@@ -231,7 +244,113 @@ def test_up_error(
...
+
+        with pytest.raises(SystemExit):
+            up(args)
+
+        # Ensure the DEVSERVICES_DEPENDENCIES_CACHE_DIR_KEY is set and is relative
+        env_vars = mock_run.call_args[1]["env"]
+        assert (
+            env_vars[DEVSERVICES_DEPENDENCIES_CACHE_DIR_KEY]
+            == f"../dependency-dir/{DEPENDENCY_CONFIG_VERSION}"
+        )
```
IanWoodard, line 300: We technically don't care about this anymore as we moved away from `include`

IanWoodard (approved):
Looks good! Left a few comments but otherwise we should be good to go

tests/commands/test_up.py
hubertdeng123: I see this referenced quite a lot throughout our codebase, I think we should clean this up separately

devservices/utils/docker.py
hubertdeng123: ticketed this

tests/utils/test_docker.py
```diff
@@ -264,15 +268,24 @@ def test_wait_for_healthy_no_healthcheck(mock_check_output: mock.Mock) -> None:
...
     mock_status = mock.Mock()
     mock_check_output.side_effect = ["unhealthy", "healthy"]
-    with mock.patch("devservices.utils.docker.HEALTHCHECK_TIMEOUT", 2), mock.patch(
-        "devservices.utils.docker.HEALTHCHECK_INTERVAL", 1
+
+    with (
+        mock.patch("devservices.utils.docker.HEALTHCHECK_TIMEOUT", HEALTHCHECK_TIMEOUT),
+        mock.patch(
+            "devservices.utils.docker.HEALTHCHECK_INTERVAL", HEALTHCHECK_INTERVAL
+        ),
```
IanWoodard, line 351: Why mock these when we can just update the frozen time with freeze gun?

tests/utils/test_docker.py
```diff
@@ -299,17 +312,27 @@ def test_wait_for_healthy_initial_check_failed_then_success(
...
     mock_status = mock.Mock()
     mock_check_output.side_effect = subprocess.CalledProcessError(1, "cmd")
     with pytest.raises(DockerError):
-        with mock.patch("devservices.utils.docker.HEALTHCHECK_TIMEOUT", 2), mock.patch(
-            "devservices.utils.docker.HEALTHCHECK_INTERVAL", 1
-        ):
+        with mock.patch(
+            "devservices.utils.docker.HEALTHCHECK_TIMEOUT", HEALTHCHECK_TIMEOUT
+        ), mock.patch(
+            "devservices.utils.docker.HEALTHCHECK_INTERVAL", HEALTHCHECK_INTERVAL
```
IanWoodard, line 399: Same as above

devservices/utils/docker_compose.py
```diff
@@ -95,6 +95,32 @@ def install_docker_compose() -> None:
...
+            text=True,
+        ).splitlines()
+    except subprocess.CalledProcessError as e:
+        raise DockerComposeError(
+            command=f"docker compose -p {project_name} -f {config_path} ps --format {{.Name}}",
+            returncode=e.returncode,
+            stdout=e.stdout,
+            stderr=e.stderr,
+        ) from e
+    return container_names
```
IanWoodard, line 121: Nit: return inside the try block since the other case raises an error which doesn't need to return

devservices/commands/up.py
```diff
@@ -144,8 +148,31 @@ def _up(
...
+        for future in concurrent.futures.as_completed(futures):
+            _ = future.result()
+
     for cmd in docker_compose_commands:
-        _bring_up_dependency(cmd, current_env, status)
+        try:
+            container_names = get_container_names_for_project(
+                cmd.project_name, cmd.config_path
+            )
+            containers_to_check.extend(container_names)
```
IanWoodard, line 165: Nit: since we don't use `container_names` anywhere else, why don't we combine these two lines?

tests/utils/test_docker.py
```diff
@@ -290,3 +297,195 @@ def test_stop_containers_remove_error(
...
+                    "-f",
+                    "{{if .State.Health}}{{.State.Health.Status}}{{else}}unknown{{end}}",
+                    "container1",
+                ],
+                stderr=subprocess.DEVNULL,
+                text=True,
+            ),
+        ]
+    )
+    mock_sleep.assert_called_once_with(5)
```
IanWoodard, line 383: Nit: shouldn't we use the constant here?

devservices/commands/up.py
hubertdeng123: I actually opted to do this here, since I think it's more readable than 

`containers_to_check.extend(get_container_names_for_project(cmd.project_name, cmd.config_path))`

tests/utils/test_docker.py
hubertdeng123: Good catch, these dont belong there

PR #179, opened 2024-12-12
ref(update): Adding caching to checking for latest version

IanWoodard:
Currently we hit the GitHub api each time a command is run which can lead to being rate limited and adds additional bloat that makes commands feel less snappy. To remedy this, this change adds a simple caching mechanism that stores the version it retrieves from the GitHub api for 15 minutes before refetching again. This should be often enough to still promptly inform users of updates while not sending unnecessary requests to the GitHub api.
I also moved `check_for_update` out of commands and into utils where I believe it belongs. I also added some preliminary testing for `check_for_update`.


devservices/utils/check_for_update.py
hubertdeng123, line 1: good call on moving this

devservices/utils/check_for_update.py
```diff
@@ -0,0 +1,75 @@
...
+
+def _delete_cached_version() -> None:
+    os.remove(DEVSERVICES_LATEST_VERSION_CACHE_FILE)
+
+
+def _get_cached_version() -> str | None:
+    if not os.path.exists(DEVSERVICES_LATEST_VERSION_CACHE_FILE):
+        return None
+    try:
+        with open(DEVSERVICES_LATEST_VERSION_CACHE_FILE, "r", encoding="utf-8") as f:
```
hubertdeng123, line 22: Thought about this a little bit more, thinking it may be better just store the latest version directly in the file. Then, we can use `os.path.getmtime` to get the timestamp the file was edited and use that instead so we don't have to perform json decoding/loading. What do you think?

IanWoodard: I felt that it would be easier for testing to have the timestamp be in the payload, but I can look into using the file itself. Let me think if there are any downsides.

IanWoodard: Fair point, updated the approach

devservices/utils/check_for_update.py
```diff
@@ -0,0 +1,59 @@
...
+
+    with urlopen(DEVSERVICES_RELEASES_URL) as response:
+        if response.status == 200:
+            data = json.loads(response.read())
+            latest_version = str(data["tag_name"])
+
+            _set_cached_version(latest_version)
+
+            return latest_version
+    return None
```
hubertdeng123, line 59: I do think we should update `main.py` to account for returning None here, but that is out of scope here as I should've done that in the past.

hubertdeng123 (approved):
Looking good! 🚀

devservices/utils/check_for_update.py
IanWoodard: we were already returning None, we can definitely do this but let's do it separately.

PR #181, opened 2024-12-17
ref(dependencies): Adding more context to sentry

IanWoodard:
We've had some intermittent issues with git fetch that are difficult to debug with the current context we have in Sentry. This change aims to improve the context we give to Sentry by including some of the git config to help us better debug future issues with git fetch.


hubertdeng123 (approved):

hubertdeng123 (commented):
looks like CI not passing, it's also passing locally for me so maybe some git config is missing here on a clean runner?

PR #183, opened 2024-12-18
feat(devservices): Return namedtuple for docker compose command

hubertdeng123:
This allows us to determine the project name, services list, and config file path of each docker compose command. It removes hacky logic from up/down to obtain dependency names, and allows us to be more flexible in the future

### DIFF
--- a/devservices/constants.py
+++ b/devservices/constants.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import os
 from datetime import timedelta
+from typing import NamedTuple
 
 MINIMUM_DOCKER_COMPOSE_VERSION = "2.29.7"
 DEVSERVICES_DIR_NAME = "devservices"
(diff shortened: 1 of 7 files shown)


devservices/constants.py
```diff
@@ -14,7 +15,6 @@
 DEVSERVICES_DEPENDENCIES_CACHE_DIR_KEY = "DEVSERVICES_DEPENDENCIES_CACHE_DIR"
 STATE_DB_FILE = os.path.join(DEVSERVICES_LOCAL_DIR, "state")
 DEVSERVICES_ORCHESTRATOR_LABEL = "orchestrator=devservices"
-DOCKER_COMPOSE_COMMAND_LENGTH = 7
```
IanWoodard, line 17: Thank goodness we can finally get rid of this 😌

devservices/constants.py
```diff
@@ -38,3 +38,10 @@
     DEVSERVICES_CACHE_DIR, "latest_version.txt"
 )
 DEVSERVICES_LATEST_VERSION_CACHE_TTL = timedelta(minutes=15)
+
+
+class DockerComposeCommand(NamedTuple):
+    full_command: list[str]
+    project_name: str
+    config_path: str
+    services: list[str]
```
IanWoodard, line 47: I'm not sure this belongs here. Shouldn't this live in `docker_compose.py`?

IanWoodard (approved):

devservices/constants.py
hubertdeng123: I will move

PR #184, opened 2024-12-19
feat(purge): Purge removes corresponding volumes

IanWoodard:
Adding functionality to remove volumes used by active devservices containers during the purge command. This aims to make purge more effective at clearing all state. One edge case that will still exist that this logic doesn't account for is clearing volumes for containers that are no longer up.

### DIFF
--- a/devservices/commands/purge.py
+++ b/devservices/commands/purge.py
@@ -56,8 +56,6 @@ def purge(_args: Namespace) -> None:
     ):
         try:
             stop_containers(devservices_containers, should_remove=True)
-        except DockerDaemonNotRunningError:
-            console.warning("The docker daemon is not running, no containers to stop")
         except DockerError as e:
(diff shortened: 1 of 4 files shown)


devservices/commands/purge.py
```diff
@@ -35,18 +37,41 @@ def purge(_args: Namespace) -> None:
...
+            stop_containers(devservices_containers, should_remove=True)
         except DockerError as e:
             console.failure(f"Failed to stop running devservices containers {e.stderr}")
             exit(1)
 
+    console.warning("Removing any devservices docker volumes")
+    if len(devservices_volumes) == 0:
+        console.success("No devservices volumes found to remove")
+    else:
+        subprocess.run(
```
hubertdeng123, line 67: Should we catch exceptions that are thrown here?

IanWoodard: Updated!

hubertdeng123 (approved):

PR #189, opened 2024-12-30
chore(devservices): Remove update warning message

hubertdeng123:
This message can be confusing, especially since in virtual environments this is not actionable for people.


IanWoodard (approved):
Is `check_for_update` used anywhere else?

hubertdeng123:
Yes, in the `devservices update` command. But, should keep that around imo

PR #201, opened 2025-01-09
fix(down): Fixing down's logic and adding proper tests

IanWoodard:
Fixes: https://getsentry.atlassian.net/browse/DEVINFRA-570 and https://getsentry.atlassian.net/browse/DEVINFRA-582.
Adds logic to properly handle bringing services down in specific edge cases that were previously not supported. This includes bringing down a service that is being used by another, which should not bring anything down. Additionally, it supports bringing down base services such as redis which are considered remote to other services but local to themselves. Additionally, this adds proper testing so we shouldn't have regressions going forward.

It is worth noting that this doesn't support what I would consider 'ideal' behavior when it comes to modes. It takes the conservative approach and doesn't bring things down unnecessarily, but doesn't stop certain dependencies based on the modes you are running. This logic would be quite difficult to add, and while we will likely add it in the future, it is not a top priority for us right now.

### DIFF
--- a/devservices/commands/down.py
+++ b/devservices/commands/down.py
@@ -101,7 +101,6 @@ def down(args: Namespace) -> None:
             status.failure(str(de))
             exit(1)
         try:
-            # TODO: What does this even mean? Why are we filtering out shared dependencies?
             remote_dependencies = get_non_shared_remote_dependencies(
                 service, remote_dependencies
             )
--- a/devservices/commands/down.py
+++ b/devservices/commands/down.py
@@ -110,7 +110,7 @@ def down(args: Namespace) -> None:
 
         # Check if any service depends on the service we are trying to bring down
         # TODO: We should also take into account the active modes of the other services (this is not trivial to do)
-        other_started_services = set(started_services) - {service.name}
+        other_started_services = set(started_services).difference({service.name})
         dependent_service_name = None
         for other_started_service in other_started_services:
             other_service = find_matching_service(other_started_service)
--- a/devservices/commands/down.py
+++ b/devservices/commands/down.py
@@ -22,6 +22,7 @@ from devservices.exceptions import DockerComposeError
 from devservices.exceptions import ServiceNotFoundError
 from devservices.utils.console import Console
 from devservices.utils.console import Status
+from devservices.utils.dependencies import construct_dependency_graph
 from devservices.utils.dependencies import get_non_shared_remote_dependencies
 from devservices.utils.dependencies import install_and_verify_dependencies
 from devservices.utils.dependencies import InstalledRemoteDependency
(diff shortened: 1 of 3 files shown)


devservices/commands/down.py
```diff
@@ -107,12 +108,32 @@ def down(args: Namespace) -> None:
...
+            dependency_graph = construct_dependency_graph(
+                other_service, other_service_active_modes
+            )
+            # If the service we are trying to bring down is in the dependency graph of another service, we should not bring it down
+            if service.name in dependency_graph.graph:
+                has_dependent_service = True
+                break
+
+        # If no other service depends on the service we are trying to bring down, we can bring it down
+        if not has_dependent_service:
```
hubertdeng123, line 130: I'm thinking it would be helpful to display a message when the current service cannot be brought down if it is used by another service. That way people won't be confused when they try bringing down redis and discovering that nothing was brought down

devservices/commands/down.py
```diff
@@ -107,12 +108,32 @@ def down(args: Namespace) -> None:
...
-        try:
-            _down(service, remote_dependencies, list(mode_dependencies), status)
-        except DockerComposeError as dce:
-            capture_exception(dce)
-            status.failure(f"Failed to stop {service.name}: {dce.stderr}")
-            exit(1)
+
+        # Check if any service depends on the service we are trying to bring down
+        # TODO: We should also take into account the active modes of the other services (this is not trivial to do)
+        other_started_services = set(started_services) - {service.name}
```
hubertdeng123, line 114: nit: we're using set notation through `.difference` down below, could we change that here as well for standardization purposes?

devservices/commands/down.py
```diff
@@ -100,19 +101,40 @@ def down(args: Namespace) -> None:
             status.failure(str(de))
             exit(1)
         try:
+            # TODO: What does this even mean? Why are we filtering out shared dependencies?
```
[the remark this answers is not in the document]
IanWoodard: Already removed

devservices/commands/down.py
IanWoodard: That's a fair point. I'll look into it

devservices/commands/down.py
```diff
@@ -100,19 +101,40 @@ def down(args: Namespace) -> None:
...
             exit(1)
-        try:
-            _down(service, remote_dependencies, list(mode_dependencies), status)
-        except DockerComposeError as dce:
-            capture_exception(dce)
-            status.failure(f"Failed to stop {service.name}: {dce.stderr}")
-            exit(1)
+
+        # Check if any service depends on the service we are trying to bring down
+        # TODO: We should also take into account the active modes of the other services (this is not trivial to do)
```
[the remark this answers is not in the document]
hubertdeng123: https://getsentry.atlassian.net/browse/DEVINFRA-586

hubertdeng123 (approved):

PR #207, opened 2025-01-13
chore(state): Refactor state to support starting services table

hubertdeng123:
This will allow us to support a `starting_services` table, allowing us to clean up intermediate states users may get into if `devservices up` fails. Then, `devservices down` will clean things up

### DIFF
--- a/devservices/utils/state.py
+++ b/devservices/utils/state.py
@@ -2,13 +2,13 @@ from __future__ import annotations
 
 import os
 import sqlite3
-from enum import StrEnum
+from enum import Enum
 
 from devservices.constants import DEVSERVICES_LOCAL_DIR
 from devservices.constants import STATE_DB_FILE
 
 
-class StateTables(StrEnum):
+class StateTables(Enum):
     STARTED_SERVICES_TABLE = "started_services"
     STARTING_SERVICES_TABLE = "starting_services"
 
--- a/devservices/utils/state.py
+++ b/devservices/utils/state.py
@@ -9,8 +9,8 @@ from devservices.constants import STATE_DB_FILE
 
 
 class StateTables(Enum):
-    STARTED_SERVICES_TABLE = "started_services"
-    STARTING_SERVICES_TABLE = "starting_services"
+    STARTED_SERVICES = "started_services"
+    STARTING_SERVICES = "starting_services"
 
 
 class State:
@@ -30,9 +30,10 @@ class State:
 
     def initialize_database(self) -> None:
         cursor = self.conn.cursor()
+        # Formatted strings here and throughout the fileshould be extremely low risk given these are constants
         cursor.execute(
             f"""
-            CREATE TABLE IF NOT EXISTS {StateTables.STARTED_SERVICES_TABLE.value} (
+            CREATE TABLE IF NOT EXISTS {StateTables.STARTED_SERVICES.value} (
                 service_name TEXT PRIMARY KEY,
                 mode TEXT,
                 timestamp TIMESTAMP DEFAULT CURRENT_TIMESTAMP
@@ -42,7 +43,7 @@ class State:
 
         cursor.execute(
             f"""
-            CREATE TABLE IF NOT EXISTS {StateTables.STARTING_SERVICES_TABLE.value} (
+            CREATE TABLE IF NOT EXISTS {StateTables.STARTING_SERVICES.value} (
                 service_name TEXT PRIMARY KEY,
                 mode TEXT,
                 timestamp TIMESTAMP DEFAULT CURRENT_TIMESTAMP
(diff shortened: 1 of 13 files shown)


devservices/utils/state.py
```diff
@@ -35,52 +51,56 @@ def initialize_database(self) -> None:
...
         self.conn.commit()
 
-    def remove_started_service(self, service_name: str) -> None:
+    def remove_service_entry(self, service_name: str, table: StateTables) -> None:
         cursor = self.conn.cursor()
         cursor.execute(
-            """
-            DELETE FROM started_services WHERE service_name = ?
+            f"""
+            DELETE FROM {table.value} WHERE service_name = ?
```
IanWoodard, line 82: We should use placeholders rather than f-strings here

devservices/utils/state.py
```diff
@@ -35,52 +51,56 @@ def initialize_database(self) -> None:
...
+                UPDATE {table.value} SET mode = ? WHERE service_name = ?
             """,
                 (",".join(active_modes + [mode]), service_name),
             )
         else:
             cursor.execute(
-                """
-                INSERT INTO started_services (service_name, mode) VALUES (?, ?)
+                f"""
+                INSERT INTO {table.value} (service_name, mode) VALUES (?, ?)
```
IanWoodard, line 72: We should use placeholders rather than f-strings here

devservices/utils/state.py
```diff
@@ -35,52 +51,56 @@ def initialize_database(self) -> None:
...
+        active_modes = self.get_active_modes_for_service(service_name, table)
+        if service_name in service_entries and mode in active_modes:
             return
-        if service_name in started_services:
+        if service_name in service_entries:
             cursor.execute(
-                """
-                UPDATE started_services SET mode = ? WHERE service_name = ?
+                f"""
+                UPDATE {table.value} SET mode = ? WHERE service_name = ?
```
IanWoodard, line 65: We should use placeholders rather than f-strings here

devservices/utils/state.py
```diff
@@ -35,52 +51,56 @@ def initialize_database(self) -> None:
...
         self.conn.commit()
 
-    def get_started_services(self) -> list[str]:
+    def get_service_entries(self, table: StateTables) -> list[str]:
         cursor = self.conn.cursor()
         cursor.execute(
-            """
-            SELECT service_name FROM started_services
+            f"""
+            SELECT service_name FROM {table.value}
```
IanWoodard, line 92: We should use placeholders rather than f-strings here

devservices/utils/state.py
```diff
@@ -35,52 +51,56 @@ def initialize_database(self) -> None:
...
-    def get_active_modes_for_service(self, service_name: str) -> list[str]:
+    def get_active_modes_for_service(
+        self, service_name: str, table: StateTables
+    ) -> list[str]:
         cursor = self.conn.cursor()
         cursor.execute(
-            """
-            SELECT mode FROM started_services WHERE service_name = ?
+            f"""
+            SELECT mode FROM {table.value} WHERE service_name = ?
```
IanWoodard, line 103: We should use placeholders rather than f-strings here

devservices/utils/state.py
```diff
@@ -92,8 +112,13 @@ def get_active_modes_for_service(self, service_name: str) -> list[str]:
...
         cursor.execute(
-            """
-            DELETE FROM started_services
+            f"""
+            DELETE FROM {StateTables.STARTED_SERVICES_TABLE.value}
+        """
+        )
+        cursor.execute(
+            f"""
+            DELETE FROM {StateTables.STARTING_SERVICES_TABLE.value}
```
IanWoodard, line 121: We should use placeholders rather than f-strings here

devservices/utils/state.py
```diff
@@ -25,8 +31,18 @@ def __new__(cls) -> State:
...
+                service_name TEXT PRIMARY KEY,
+                mode TEXT,
+                timestamp TIMESTAMP DEFAULT CURRENT_TIMESTAMP
+            )
+        """
+        )
+
+        cursor.execute(
+            f"""
+            CREATE TABLE IF NOT EXISTS {StateTables.STARTING_SERVICES_TABLE.value} (
```
IanWoodard, line 45: We should use placeholders rather than f-strings here

devservices/utils/state.py
```diff
@@ -25,8 +31,18 @@ def __new__(cls) -> State:
     def initialize_database(self) -> None:
         cursor = self.conn.cursor()
         cursor.execute(
-            """
-            CREATE TABLE IF NOT EXISTS started_services (
+            f"""
+            CREATE TABLE IF NOT EXISTS {StateTables.STARTED_SERVICES_TABLE.value} (
```
IanWoodard, line 35: We should use placeholders rather than f-strings here

hubertdeng123: placeholders do not work for table names and column names, so we will need to keep as is

these are constants defined at the top of the file, not user input so not risky

IanWoodard (approved):

PR #212, opened 2025-01-15
ref(commands): Improving commands to use starting services and started

IanWoodard:
https://getsentry.atlassian.net/browse/DEVINFRA-596 Now that we track both starting and started services, commands such as `logs` and `list-services` should be able to work with starting services. This change updates those commands to do just that by treating both starting and started services as not-stopped (aka running).

### DIFF
--- a/devservices/commands/list_services.py
+++ b/devservices/commands/list_services.py
@@ -31,7 +31,9 @@ def list_services(args: Namespace) -> None:
     coderoot = get_coderoot()
     services = get_local_services(coderoot)
     state = State()
-    running_services = state.get_service_entries(StateTables.STARTED_SERVICES)
+    starting_services = set(state.get_service_entries(StateTables.STARTING_SERVICES))
+    started_services = set(state.get_service_entries(StateTables.STARTED_SERVICES))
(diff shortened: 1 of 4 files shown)


devservices/commands/list_services.py
```diff
@@ -31,7 +31,9 @@ def list_services(args: Namespace) -> None:
     coderoot = get_coderoot()
     services = get_local_services(coderoot)
     state = State()
-    running_services = state.get_service_entries(StateTables.STARTED_SERVICES)
+    starting_services = set(state.get_service_entries(StateTables.STARTING_SERVICES))
+    started_services = set(state.get_service_entries(StateTables.STARTED_SERVICES))
+    running_services = starting_services.union(started_services)
```
hubertdeng123, line 36: nit: I think it is a bit confusing that we're using `running_services` here as a variable name but using `running` in list_services to indicate a service has been `started`. I think we should perhaps change 

```
elif service.name in started_services:
            status = "running"
```
to
```
elif service.name in started_services:
            status = "started"
```

hubertdeng123 (approved):

PR #215, opened 2025-01-16
ref(dependencies): Improving error handling for dependency logic

IanWoodard:
https://getsentry.atlassian.net/browse/DEVINFRA-598. Adding additional error handling to the core dependency logic. Fixed the `_run_command` function to now properly propagate the stderr which it did not do previously which should now give us more information into the failures we see. For now I opted not to add retries to anything besides fetch since fetch is by far the most likely to face transient errors, and retrying certain other commands can put us in a potentially problematic state.
More improvements to the core dependency logic, specifically around the git config are coming in another PR.

### DIFF
--- a/devservices/utils/dependencies.py
+++ b/devservices/utils/dependencies.py
@@ -485,24 +485,50 @@ def _update_dependency(
         ) from e
 
     # Check if the local repo is up-to-date
-    local_commit = subprocess.check_output(
-        ["git", "rev-parse", "HEAD"],
-        cwd=dependency_repo_dir,
-        stderr=subprocess.PIPE,
-    ).strip()
+    try:
+        local_commit = subprocess.check_output(
+            ["git", "rev-parse", "HEAD"],
+            cwd=dependency_repo_dir,
+            stderr=subprocess.PIPE,
+        ).strip()
+    except subprocess.CalledProcessError as e:
+        raise DependencyError(
+            repo_name=dependency.repo_name,
+            repo_link=dependency.repo_link,
+            branch=dependency.branch,
+        ) from e
 
-    remote_commit = subprocess.check_output(
-        ["git", "rev-parse", "FETCH_HEAD"],
-        cwd=dependency_repo_dir,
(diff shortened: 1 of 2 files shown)


devservices/utils/dependencies.py
```diff
@@ -589,12 +614,22 @@ def _has_remote_config(remote_config: RemoteConfig | None) -> TypeGuard[RemoteCo
     return remote_config is not None
 
 
+def _rev_parse(repo_dir: str, ref: str) -> str:
+    return (
```
hubertdeng123, line 618: should we add debug logs here?

devservices/utils/dependencies.py
```diff
@@ -485,24 +485,42 @@ def _update_dependency(
...
-        cwd=dependency_repo_dir,
-        stderr=subprocess.PIPE,
-    ).strip()
-
-    remote_commit = subprocess.check_output(
-        ["git", "rev-parse", "FETCH_HEAD"],
-        cwd=dependency_repo_dir,
-        stderr=subprocess.PIPE,
-    ).strip()
+    try:
```
hubertdeng123, line 488: One concern I have here is that the error message is not detailed enough. Previously, this would at least give a bit more context but now it is displayed as
`DependencyError: snuba (https://github.com/getsentry/snuba.git) on master`

I think some more context would be welcome here

devservices/utils/dependencies.py
hubertdeng123: also I do realize these debug logs were not there before, but thought they might be nice to add

devservices/utils/dependencies.py
IanWoodard: Previously we didn't have any error handling here, but we can definitely add better error messages, I plan on doing that in the next PR which is coming soon.

devservices/utils/dependencies.py
IanWoodard: We could

hubertdeng123 (approved):

PR #219, opened 2025-01-24
feat(reset): Adding reset command

IanWoodard:
https://getsentry.atlassian.net/browse/DEVINFRA-604
Adding the ability to reset specific docker volumes.


devservices/utils/docker.py
```diff
@@ -186,7 +188,7 @@ def stop_containers(containers: list[str], should_remove: bool = False) -> None:
             ["docker", "stop"] + containers,
             check=True,
             stdout=subprocess.DEVNULL,
-            stderr=subprocess.DEVNULL,
+            stderr=subprocess.PIPE,
```
hubertdeng123, line 191: Previously, stderr was always None because this wasn't being passed in

devservices/commands/reset.py
```diff
@@ -0,0 +1,110 @@
...
+                DEVSERVICES_ORCHESTRATOR_LABEL,
+                f"com.docker.compose.service={args.service_name}",
+            ]
+        )
+    except DockerDaemonNotRunningError as e:
+        console.warning(str(e))
+        return
+    except DockerError as e:
+        console.failure(f"Failed to get matching containers {e.stderr}")
+        exit(1)
```
hubertdeng123, line 53: should not use exit(1) here and below to avoid transaction failure

devservices/commands/reset.py
```diff
@@ -0,0 +1,110 @@
...
+    if len(matching_volumes) == 0:
+        console.failure(f"No volumes found for {service_name}")
+        exit(1)
+
+    state = State()
+    started_services = set(state.get_service_entries(StateTables.STARTED_SERVICES))
+    starting_services = set(state.get_service_entries(StateTables.STARTING_SERVICES))
+    active_service_names = starting_services.union(started_services)
+
+    # TODO: Should we add threading here to speed up the process?
```
hubertdeng123, line 74: Perhaps, but seems fine as is. Let's change this to a statement not a question

devservices/commands/reset.py
hubertdeng123: actually i take that back, this is fine

hubertdeng123 (approved):

PR #224, opened 2025-01-29
ref(sentry): Tweak devservices environment env variable

hubertdeng123:
Making this a string instead of a boolean specifically so that runners can report to Sentry under the ci environment

### DIFF
--- a/devservices/main.py
+++ b/devservices/main.py
@@ -32,8 +32,8 @@ from devservices.utils.console import Console
 from devservices.utils.docker_compose import check_docker_compose_version
 from devservices.utils.git import get_git_version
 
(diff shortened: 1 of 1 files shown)


IanWoodard (approved):
Besides our local environments, is there anywhere else we used this that we may need to update? Also, as we talked about earlier, we should set the environment to CI in CI.

hubertdeng123:
Actually, I think a better idea here is to check for the environment variable CI here

IanWoodard (approved):
This is also a reasonable approach, I did kind of like the ability to specify any environment, but this will likely be better for consistency.

PR #234, opened 2025-02-13
feat(readme): Add config section to readme

hubertdeng123:
This adds a lot more description for folks on how to setup a devservices config

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ The recommended way to install devservices is through a virtualenv in the requir
 devservices==1.0.16
 ```
 
-### 2. Add Devservices Config Files
+### 2. Add devservices config files
 
 Each repo should have a `devservices` directory with a `config.yml` file. This file is used to define services, dependencies, and modes. Other files and subdirectories in the `devservices` directory are optional and can be most commonly used for volume mounts.
 
--- a/README.md
+++ b/README.md
@@ -31,3 +31,114 @@ NOTE: service-name is an optional parameter. If not provided, devservices will a
 - `devservices list-dependencies <service-name>`: List all dependencies for a service and whether they are enabled/disabled.
 - `devservices update` Update devservices to the latest version.
 - `devservices purge`: Purge the local devservices cache.
+
+## Configuration
+
+devservices uses a configuration file to define services, dependencies, and modes. The configuration file is located in the `devservices` directory of your repository.
+
+The configuration file is a yaml file that looks like this:
+
+```yaml
+# This is a yaml block that holds devservices specific configuration settings. This is comprised of a few main sections:
+# - version: The version of the devservices config file. This is used to ensure compatibility between devservices and the config file.
+# - service_name: The name of the service. This is used to identify the service in the config file.
+# - dependencies: A list of dependencies for the service. Each dependency is a yaml block that holds the dependency configuration. There are two types of dependencies:
+#   - local: A dependency that is defined in the config file. These dependencies do not have a remote field.
+#   - remote: A dependency that is defined in the devservices directory in a remote repository. These configs are automatically fetched from the remote repository and installed. Any dependency with a remote field will be treated as a remote dependency.
+# - modes: A list of modes for the service. Each mode includes a list of dependencies that are enabled in that mode.
+x-sentry-service-config:
+  version: 0.1
+  service_name: example-service
+  dependencies:
+    example-dependency-1:
+      description: Example dependency defined in the config file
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -31,3 +22,122 @@ NOTE: service-name is an optional parameter. If not provided, devservices will a
...
+The configuration file is a yaml file that looks like this:
+
+```yaml
+# This is a yaml block that holds devservices specific configuration settings. This is comprised of a few main sections:
+# - version: The version of the devservices config file. This is used to ensure compatibility between devservices and the config file.
+# - service_name: The name of the service. This is used to identify the service in the config file.
+# - dependencies: A list of dependencies for the service. Each dependency is a yaml block that holds the dependency configuration. There are two types of dependencies:
+#   - local: A dependency that is defined in the config file. These dependencies do not have a remote field.
+#   - remote: A dependency that is defined in the devservices directory in a remote repository. These configs are automatically fetched from the remote repository and installed. Any dependency with a remote field will be treated as a remote dependency. Example: https://github.com/getsentry/snuba/blob/59a5258ccbb502827ebc1d3b1bf80c607a3301bf/devservices/config.yml#L8
+# - modes: A list of modes for the service. Each mode includes a list of dependencies that are enabled in that mode.
```
IanWoodard, line 49: Nit:
```suggestion
# - modes: A list of modes for the service. Each mode includes a list of dependencies that are used in that mode.
```

README.md
```diff
@@ -31,3 +22,122 @@ NOTE: service-name is an optional parameter. If not provided, devservices will a
...
+Each repo should have a `devservices` directory with a `config.yml` file. This file is used to define services, dependencies, and modes. Other files and subdirectories in the `devservices` directory are optional and can be most commonly used for volume mounts.
+
+The configuration file is a yaml file that looks like this:
+
+```yaml
+# This is a yaml block that holds devservices specific configuration settings. This is comprised of a few main sections:
+# - version: The version of the devservices config file. This is used to ensure compatibility between devservices and the config file.
+# - service_name: The name of the service. This is used to identify the service in the config file.
+# - dependencies: A list of dependencies for the service. Each dependency is a yaml block that holds the dependency configuration. There are two types of dependencies:
+#   - local: A dependency that is defined in the config file. These dependencies do not have a remote field.
```
IanWoodard, line 47: Should we provide more information about why you may want to use this versus a remote dependency?

README.md
```diff
@@ -31,3 +22,122 @@ NOTE: service-name is an optional parameter. If not provided, devservices will a
...
+### 2. Add devservices config files
+
+Each repo should have a `devservices` directory with a `config.yml` file. This file is used to define services, dependencies, and modes. Other files and subdirectories in the `devservices` directory are optional and can be most commonly used for volume mounts.
+
+The configuration file is a yaml file that looks like this:
+
+```yaml
+# This is a yaml block that holds devservices specific configuration settings. This is comprised of a few main sections:
+# - version: The version of the devservices config file. This is used to ensure compatibility between devservices and the config file.
+# - service_name: The name of the service. This is used to identify the service in the config file.
```
IanWoodard, line 45: I forget, do we have any guidelines on what we expect people to name the service in accordance with the repo?

README.md
```diff
@@ -31,3 +22,122 @@ NOTE: service-name is an optional parameter. If not provided, devservices will a
...
+      timeout: 5s
+      retries: 3
+    environment:
+      EXAMPLE_ENV_VAR: example-value
+    extra_hosts:
+      host.docker.internal: host-gateway
+    networks:
+      - devservices
+    labels:
+      - orchestrator=devservices
```
IanWoodard, line 129: We should make clear this is strictly required

IanWoodard: Maybe even validate that all configs include these

IanWoodard: I see you do mention it is required above, but could we make it even clearer somehow?

IanWoodard (commented):
Some initial feedback, looking good so far

README.md
hubertdeng123: yeah, i can add a comment inline reminding people that anything below a certain line is required

README.md
hubertdeng123: we do not

README.md
hubertdeng123: Done, added a sentence

README.md
hubertdeng123: done

IanWoodard (approved):

PR #237, opened 2025-02-19
feat(sentry): Manually set status of transactions

hubertdeng123:
For transactions that do not have a trace that is tied to an error that has a higher level than "info", we manually set the status to unknown

### DIFF
--- a/devservices/main.py
+++ b/devservices/main.py
@@ -47,15 +47,27 @@ current_version = metadata.version("devservices")
 error_trace_ids = set()
 
 
-# This is a hack to get the trace_id from the errors we care about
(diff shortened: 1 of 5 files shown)


devservices/main.py
```diff
@@ -50,6 +69,8 @@
         enable_tracing=True,
         integrations=[ArgvIntegration()],
         environment=sentry_environment,
+        before_send=set_error_status,
+        before_send_transaction=set_transaction_status,
```
IanWoodard, line 73: Could we just call these before_send_error and before_send_transaction since while we currently use these for specific purposes, that may not be the case later and it would be good to keep the naming open-ended.

IanWoodard (approved):

PR #240, opened 2025-03-11
feat(runtime): Updating up to handle local runtime

IanWoodard:
Updating the up command to not bring up dependencies that are set to use a local runtime.
https://getsentry.atlassian.net/browse/DEVINFRA-638

### DIFF
--- a/tests/commands/test_up.py
+++ b/tests/commands/test_up.py
@@ -17,6 +17,8 @@ from devservices.exceptions import ContainerHealthcheckFailedError
 from devservices.exceptions import DependencyError
 from devservices.exceptions import DockerComposeError
 from devservices.exceptions import ServiceNotFoundError
+from devservices.utils.docker_compose import DockerComposeCommand
+from devservices.utils.state import ServiceRuntime
 from devservices.utils.state import State
 from devservices.utils.state import StateTables
 from testing.utils import create_config_file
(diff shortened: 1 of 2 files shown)


tests/commands/test_up.py
```diff
@@ -1018,3 +1020,154 @@ def test_up_service_not_found_error(
...
+            "services": {
+                "redis": {"image": "redis:6.2.14-alpine"},
+            },
+        }
+        create_config_file(redis_repo_path, redis_config)
+        run_git_command(["add", "."], cwd=redis_repo_path)
+        run_git_command(["commit", "-m", "Add devservices config"], cwd=redis_repo_path)
+
+        example_repo_path = create_mock_git_repo(
+            "blank_repo", tmp_path / "example-service"
```
hubertdeng123, line 1067: for clarity I think we should rename this to `local-runtime-service`

hubertdeng123: Could we remove redis as a dependency from the local service for the test case? Just so it's a bit easier to verify the behavior is what we want

IanWoodard: I wanted to show it brings up something still but doesn't bring up the local-runtime-service

hubertdeng123 (approved):

PR #241, opened 2025-03-11
feat(ci): Add tags for ci runs

hubertdeng123:
This adds a lot of context for devservices transactions in CI

https://getsentry.atlassian.net/browse/DEVINFRA-643

### DIFF
--- a/devservices/main.py
+++ b/devservices/main.py
@@ -11,6 +11,7 @@ from importlib import metadata
 from sentry_sdk import capture_exception
 from sentry_sdk import flush
 from sentry_sdk import init
+from sentry_sdk import set_context
 from sentry_sdk import set_tag
 from sentry_sdk import set_user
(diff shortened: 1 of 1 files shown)


devservices/main.py
```diff
@@ -84,6 +85,22 @@ def before_send_transaction(event: Event, hint: Hint) -> Event:
     username = getpass.getuser()
     set_user({"username": username})
     set_tag("user_platform", platform.platform())
+    if sentry_environment == "CI":
+        set_context(
```
IanWoodard, line 89: Is set_context okay with values being none since they can technically be none?

IanWoodard (approved):

devservices/main.py
hubertdeng123: I don't see why not, that would still be a valid object

PR #242, opened 2025-03-13
feat(up): Add better UX when pulling images

hubertdeng123:
https://github.com/user-attachments/assets/7bc19aa7-084b-4d07-b4cb-79df1bc1f8c6

### DIFF
--- a/devservices/commands/up.py
+++ b/devservices/commands/up.py
@@ -170,9 +170,9 @@ def _up(
         mode_dependencies=mode_dependencies,
     )
 
-    with concurrent.futures.ThreadPoolExecutor() as dependency_executor:
+    with concurrent.futures.ThreadPoolExecutor() as pull_dependency_executor:
         futures = [
-            dependency_executor.submit(
+            pull_dependency_executor.submit(
                 _pull_dependency_images, cmd, current_env, status
             )
             for cmd in pull_commands
@@ -192,9 +192,11 @@ def _up(
     )
 
     containers_to_check = []
-    with concurrent.futures.ThreadPoolExecutor() as dependency_executor:
+    with concurrent.futures.ThreadPoolExecutor() as up_dependency_executor:
         futures = [
-            dependency_executor.submit(_bring_up_dependency, cmd, current_env, status)
+            up_dependency_executor.submit(
+                _bring_up_dependency, cmd, current_env, status
+            )
             for cmd in up_commands
         ]
         for future in concurrent.futures.as_completed(futures):
--- a/devservices/commands/up.py
+++ b/devservices/commands/up.py
@@ -118,12 +118,20 @@ def up(args: Namespace) -> None:
     state.update_service_entry(service.name, mode, StateTables.STARTED_SERVICES)
 
 
+def _pull_dependency_images(
+    cmd: DockerComposeCommand, current_env: dict[str, str], status: Status
+) -> None:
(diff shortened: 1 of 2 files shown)


tests/commands/test_up.py
```diff
@@ -29,11 +29,11 @@
 @mock.patch("devservices.commands.up._create_devservices_network")
 @mock.patch("devservices.commands.up.check_all_containers_healthy")
 @mock.patch(
-    "devservices.commands.up.subprocess.check_output",
```
hubertdeng123, line 32: decided to be more specific in this mock throughout this file

tests/commands/test_up.py
```diff
@@ -632,36 +630,32 @@ def test_up_mode_simple(
...
-                    env=mock.ANY,
+                    mock.ANY,
                 ),
             ]
         )
+        mock_get_container_names_for_project.assert_called_once()
+
+        mock_create_devservices_network.assert_called_once()
+
+        mock_get_non_remote_services.assert_called()
```
IanWoodard, line 658: Nit: We should ideally be more precise with what we expect these to be called with (especially the last one)

tests/commands/test_up.py
```diff
@@ -506,43 +503,44 @@ def test_up_docker_compose_container_healthcheck_failed(
...
```
IanWoodard, line 543: Same as below

tests/commands/test_up.py
```diff
@@ -382,43 +382,44 @@ def test_up_docker_compose_container_lookup_error(
...
```
IanWoodard, line 422: Same as below

tests/commands/test_up.py
```diff
@@ -84,44 +84,44 @@ def test_up_simple(
...
```
IanWoodard, line 124: Same as below

devservices/commands/up.py
```diff
@@ -150,12 +158,35 @@ def _up(
...
+        service=service,
+        remote_dependencies=sorted_remote_dependencies,
+        current_env=current_env,
+        command="pull",
+        options=[],
+        service_config_file_path=service_config_file_path,
+        mode_dependencies=mode_dependencies,
+    )
+
+    with concurrent.futures.ThreadPoolExecutor() as dependency_executor:
```
IanWoodard, line 173: Nit: We should probably give it a more descriptive name and avoid re-using the same name between this and the one below.

devservices/commands/up.py
```diff
@@ -118,12 +118,20 @@ def up(args: Namespace) -> None:
     state.update_service_entry(service.name, mode, StateTables.STARTED_SERVICES)
 
 
+def _pull_dependency_images(
+    cmd: DockerComposeCommand, current_env: dict[str, str], status: Status
+) -> None:
+    run_cmd(cmd.full_command, current_env)
+    for dependency in cmd.services:
+        status.info(f"Pulled image for {dependency}")
```
IanWoodard, line 126: I think it would be nice to show them when the pulling starts so it's clearer which image may be holding things up, but I realize that would cause a fair bit more noise so I am fine with leaving it out for now.

IanWoodard (approved):

devservices/commands/up.py
hubertdeng123: I thought about this too, but it turned out to be super verbose

PR #243, opened 2025-03-14
ref(down): Updating down to work with local runtime

IanWoodard:
Updating the down command to work with services set to use a local runtime. Ensures down doesn't bring down services that are set to run locally when bringing down a dependent service.
https://getsentry.atlassian.net/browse/DEVINFRA-637

### DIFF
--- a/devservices/commands/down.py
+++ b/devservices/commands/down.py
@@ -30,6 +30,7 @@ from devservices.utils.docker_compose import get_docker_compose_commands_to_run
 from devservices.utils.docker_compose import run_cmd
 from devservices.utils.services import find_matching_service
 from devservices.utils.services import Service
+from devservices.utils.state import ServiceRuntime
 from devservices.utils.state import State
 from devservices.utils.state import StateTables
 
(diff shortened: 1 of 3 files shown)


devservices/commands/down.py
```diff
@@ -122,26 +123,16 @@ def down(args: Namespace) -> None:
...
-            )
-            other_service_active_modes = (
-                other_service_active_starting_modes
-                or other_service_active_started_modes
-            )
-            dependency_graph = construct_dependency_graph(
-                other_service, other_service_active_modes
+        # We can ignore checking if anything relies on the service
+        # if it is a locally running service
+        if service.name not in services_with_local_runtime:
```
hubertdeng123, line 132: Won't this stop us from bringing down `snuba` if it is toggled as local? That isn't intended behavior right?

hubertdeng123: nvm, read this as `service.name in services_with_local_runtime`

IanWoodard: This shouldn't stop us from bringing down snuba since we won't check if anything relies on it if it has a local runtime

tests/commands/test_down.py
```diff
@@ -968,3 +969,155 @@ def test_down_overlapping_non_remote_services(
...
+
+        os.chdir(other_service_path)
+
+        state = State()
+        state.update_service_entry(
+            "other-service", "default", StateTables.STARTED_SERVICES
+        )
+        state.update_service_runtime("local-runtime-service", ServiceRuntime.LOCAL)
+
+        args = Namespace(service_name=None, debug=False)
```
hubertdeng123, line 1087: is service_name supposed to be None here?

hubertdeng123: yes it is, i am forgetting things

IanWoodard: Yep, since I am in the `other_service_path` directory meaning it will default to using `other-service`

tests/commands/test_down.py
```diff
@@ -968,3 +969,155 @@ def test_down_overlapping_non_remote_services(
                 mock.call("example-service", StateTables.STARTED_SERVICES),
             ]
         )
+
+
+@mock.patch("devservices.utils.state.State.remove_service_entry")
+def test_down_does_not_bring_down_dependency_if_set_to_local(
```
hubertdeng123, line 975: Can we add another test case where we bring down a service that is toggled as local, but is a dependent service of another parent service? That should be allowed

IanWoodard: Added. Might add another too

IanWoodard: I wanted to show that despite the local runtime dependency having a shared dependency with the dependent service (other-service), it still respects that shared dependency and only brings down non-shared dependencies.

PR #247, opened 2025-03-28
fix(devservices):  Fix healthcheck bug

hubertdeng123:
https://linear.app/getsentry/issue/DI-660/healthcheck-bug-for-when-additional-containers-are-already-up

### DIFF
--- a/devservices/commands/up.py
+++ b/devservices/commands/up.py
@@ -218,7 +218,7 @@ def _up(
     for cmd in up_commands:
         try:
             container_names = get_container_names_for_project(
-                cmd.project_name, cmd.config_path
+                cmd.project_name, cmd.config_path, cmd.services
             )
             containers_to_check.extend(container_names)
         except DockerComposeError as dce:
--- a/devservices/utils/docker_compose.py
+++ b/devservices/utils/docker_compose.py
@@ -95,7 +95,7 @@ def install_docker_compose() -> None:
 
 
 def get_container_names_for_project(
-    project_name: str, config_path: str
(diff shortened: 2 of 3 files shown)


devservices/commands/up.py
```diff
@@ -218,7 +218,7 @@ def _up(
     for cmd in up_commands:
         try:
             container_names = get_container_names_for_project(
-                cmd.project_name, cmd.config_path
+                cmd.project_name, cmd.config_path, cmd.services
```
IanWoodard, line 221: Is there a way to make it clearer that these are the services in the mode?

devservices/utils/docker_compose.py
```diff
@@ -95,7 +95,7 @@ def install_docker_compose() -> None:
 
 
 def get_container_names_for_project(
-    project_name: str, config_path: str
+    project_name: str, config_path: str, services: list[str]
```
IanWoodard, line 98: Shouldn't this be a set in theory? I know it would make everything more difficult though

IanWoodard (approved):

devservices/commands/up.py
hubertdeng123: Not really besides naming, but I think this should be satisfactory for now, since these are the services that are involved in the docker compose cmd

devservices/utils/docker_compose.py
hubertdeng123: Yeah, that would be more accurate although this shouldn't be a problem here since container names are always unique

PR #248, opened 2025-03-28
ref(dependency-graph): Adding more context to the dependency graph

IanWoodard:
This change adds context about whether a given node in the dependency graph is a docker compose service (leaf) or a devservices service. This enables us to avoid cycles when we have docker compose services with the same name as the devservices service which is a common thing we see such as with snuba and it's container.
https://linear.app/getsentry/issue/DI-661/add-more-context-to-dependency-graph
This is required to enable the improvements we are looking to make to the status command: https://linear.app/getsentry/issue/DI-647/rework-status-command.

### DIFF
--- a/devservices/utils/dependencies.py
+++ b/devservices/utils/dependencies.py
@@ -10,6 +10,7 @@ from collections import deque
 from concurrent.futures import as_completed
 from concurrent.futures import ThreadPoolExecutor
 from dataclasses import dataclass
+from enum import Enum
 from typing import TextIO
 from typing import TypeGuard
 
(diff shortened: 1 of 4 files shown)


devservices/utils/dependencies.py
```diff
@@ -53,59 +54,73 @@
...
+
+@dataclass(frozen=True, eq=True)
+class DependencyNode:
+    name: str
+    dependency_type: DependencyType
+
+    def __str__(self) -> str:
+        return self.name
+
+    def qualified_name(self) -> str:
```
hubertdeng123, line 70: I'm not seeing this being used anywhere so wondering if it is needed

devservices/utils/dependencies.py
```diff
@@ -53,59 +54,73 @@
...
 
-    def add_edge(self, service_name: str, dependency_name: str) -> None:
-        # TODO: We should rename services that depend on themselves
-        if service_name == dependency_name:
-            return
-        if service_name not in self.graph:
-            self.add_dependency(service_name)
-        if dependency_name not in self.graph:
-            self.add_dependency(dependency_name)
+    def add_edge(self, from_node: DependencyNode, to_node: DependencyNode) -> None:
```
hubertdeng123, line 82: Do we still need to care about a case where `from_node == to_node`?

IanWoodard: In a sense, no. It would be a cycle but we have cycle detection in the topological sort function. With that said, we could add an explicit guard against it here as well. We shouldn't handle it the same way we did before though since that was a hack

devservices/utils/dependencies.py
IanWoodard: Good catch, removed it's usage earlier

hubertdeng123 (approved):
great tests 💯

PR #249, opened 2025-04-02
ref(up): Add info about local runtime in up

IanWoodard:
Adds visibility into why certain dependencies are skipped if they have a local runtime when running the up command.
https://linear.app/getsentry/issue/DI-605/improve-visibility-of-service-runtimes-for-users

### DIFF
--- a/devservices/commands/up.py
+++ b/devservices/commands/up.py
@@ -93,7 +93,8 @@ def up(args: Namespace) -> None:
         )
         for service_with_local_runtime in services_with_local_runtime:
             if (
-                service_with_local_runtime != service.name
+                mode in modes
+                and service_with_local_runtime != service.name
                 and service_with_local_runtime in modes[mode]
             ):
                 status.warning(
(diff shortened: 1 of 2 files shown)


devservices/commands/up.py
```diff
@@ -88,6 +88,17 @@ def up(args: Namespace) -> None:
         lambda: console.warning(f"Starting '{service.name}' in mode: '{mode}'"),
         lambda: console.success(f"{service.name} started"),
     ) as status:
+        services_with_local_runtime = state.get_services_by_runtime(
+            ServiceRuntime.LOCAL
+        )
+        for service_with_local_runtime in services_with_local_runtime:
+            if (
+                service_with_local_runtime != service.name
+                and service_with_local_runtime in modes[mode]
```
hubertdeng123, line 97: Do we care about the case where a service is toggled that is not in `modes[mode]`?

hubertdeng123: If not, it is something that we should at least ticket so we can track for later if it is not worth tackling in the near term

IanWoodard: Good point, will address

hubertdeng123: just to confirm: are we tracking this for later then?

hubertdeng123: nvm, I see it addressed below

tests/commands/test_up.py
```diff
@@ -1194,14 +1256,577 @@ def test_up_does_not_bring_up_dependency_if_set_to_local(
...
+
+        captured = capsys.readouterr()
+        assert (
+            "Skipping 'local-runtime-service' as it is set to run locally"
+            in captured.out.strip()
+        )
+
+
+@mock.patch("devservices.utils.state.State.update_service_entry")
+def test_up_does_not_bring_up_nested_dependency_if_set_to_local_and_mode_does_not_contain_it(
```
hubertdeng123, line 1503: nice test case!

hubertdeng123 (approved):

PR #252, opened 2025-04-11
feat(supervisor): Generate supervisor config file to local dir

hubertdeng123:
This is important because it abstracts config settings that are not relevant away from the end user and makes it easier to handle clashing that can happen between connections with ports.

This opens socket connections through sock files that are named after the devservices services. This prevents clashing ports entirely

### DIFF
--- a/devservices/constants.py
+++ b/devservices/constants.py
@@ -25,6 +25,7 @@ DEVSERVICES_CACHE_DIR = os.path.expanduser("~/.cache/sentry-devservices")
 DEVSERVICES_LOCAL_DIR = os.path.expanduser("~/.local/share/sentry-devservices")
 DEVSERVICES_DEPENDENCIES_CACHE_DIR = os.path.join(DEVSERVICES_CACHE_DIR, "dependencies")
 DEVSERVICES_DEPENDENCIES_CACHE_DIR_KEY = "DEVSERVICES_DEPENDENCIES_CACHE_DIR"
+DEVSERVICES_SUPERVISOR_CONFIG_DIR = os.path.join(DEVSERVICES_LOCAL_DIR, "supervisor")
 STATE_DB_FILE = os.path.join(DEVSERVICES_LOCAL_DIR, "state")
 DEVSERVICES_ORCHESTRATOR_LABEL = "orchestrator=devservices"
 
--- a/devservices/utils/supervisor.py
+++ b/devservices/utils/supervisor.py
@@ -59,16 +59,20 @@ class SupervisorManager:
         config.read(config_file_path)
         os.makedirs(DEVSERVICES_SUPERVISOR_CONFIG_DIR, exist_ok=True)
 
+        # Set unix http server to use the socket path
         config["unix_http_server"] = {"file": self.socket_path}
 
+        # Set generated pidfile to use service name
         config["supervisord"] = {
             "pidfile": os.path.join(
                 DEVSERVICES_SUPERVISOR_CONFIG_DIR, f"{self.service_name}.pid"
             )
         }
 
+        # Set supervisorctl to use the socket path
         config["supervisorctl"] = {"serverurl": f"unix://{self.socket_path}"}
 
+        # Required by supervisor to work properly
         config["rpcinterface:supervisor"] = {
             "supervisor.rpcinterface_factory": "supervisor.rpcinterface:make_main_rpcinterface"
         }
--- a/devservices/utils/supervisor.py
+++ b/devservices/utils/supervisor.py
@@ -1,29 +1,96 @@
 from __future__ import annotations
 
+import configparser
+import http.client
 import os
+import socket
 import subprocess
 import xmlrpc.client
+from pathlib import Path
 
+from devservices.constants import DEVSERVICES_SUPERVISOR_CONFIG_DIR
 from devservices.exceptions import SupervisorConfigError
 from devservices.exceptions import SupervisorConnectionError
 from devservices.exceptions import SupervisorError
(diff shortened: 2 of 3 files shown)


devservices/constants.py
```diff
@@ -25,6 +25,7 @@ class Color:
 DEVSERVICES_LOCAL_DIR = os.path.expanduser("~/.local/share/sentry-devservices")
 DEVSERVICES_DEPENDENCIES_CACHE_DIR = os.path.join(DEVSERVICES_CACHE_DIR, "dependencies")
 DEVSERVICES_DEPENDENCIES_CACHE_DIR_KEY = "DEVSERVICES_DEPENDENCIES_CACHE_DIR"
+DEVSERVICES_SUPERVISOR_CONFIG_DIR = os.path.join(DEVSERVICES_LOCAL_DIR, "supervisor")
```
hubertdeng123, line 28: using local dir here because if these files are deleted while devservices is running processes, unexpected behavior can occur

devservices/utils/supervisor.py
```diff
@@ -1,29 +1,96 @@
...
+        return UnixSocketHTTPConnection(self.socket_path)
+
+
 class SupervisorManager:
-    def __init__(self, port: int | None, config_file: str | None = None) -> None:
-        if port is None:
-            raise SupervisorConfigError("Port is required")
-        if config_file is None or not os.path.exists(config_file):
-            raise SupervisorConfigError("Supervisor config file not provided")
+    def __init__(self, config_file: str, service_name: str) -> None:
```
IanWoodard, line 44: Nit: could we call it `config_file_path`? That would be clearer imo

devservices/utils/supervisor.py
```diff
@@ -1,29 +1,96 @@
...
-        self.config_file = config_file
-        self.port = port
+        config = configparser.ConfigParser()
+
+        config.read(config_file)
+        os.makedirs(DEVSERVICES_SUPERVISOR_CONFIG_DIR, exist_ok=True)
+
+        self.socket_path = os.path.join(
+            DEVSERVICES_SUPERVISOR_CONFIG_DIR, f"{self.service_name}.sock"
+        )
```
IanWoodard, line 60: Could we move this to the constructor as I don't think it belongs in here

devservices/utils/supervisor.py
IanWoodard, line 44: Nit: could we call it `config_file_path`? That would be clearer imo

devservices/utils/supervisor.py
```diff
@@ -1,29 +1,96 @@
...
+
+        config["supervisorctl"] = {"serverurl": f"unix://{self.socket_path}"}
+
+        config["rpcinterface:supervisor"] = {
+            "supervisor.rpcinterface_factory": "supervisor.rpcinterface:make_main_rpcinterface"
+        }
+
+        new_config_file = os.path.join(
+            DEVSERVICES_SUPERVISOR_CONFIG_DIR, f"{self.service_name}.processes.conf"
+        )
```
IanWoodard, line 81: We should also make this clear that it is a path

devservices/utils/supervisor.py
```diff
@@ -1,29 +1,96 @@
...
+        with open(new_config_file, "w") as f:
+            config.write(f)
+
+        return str(new_config_file)
```
IanWoodard, line 85: The str is redundant here

devservices/utils/supervisor.py
```diff
@@ -1,29 +1,96 @@
...
-    def __init__(self, port: int | None, config_file: str | None = None) -> None:
-        if port is None:
-            raise SupervisorConfigError("Port is required")
-        if config_file is None or not os.path.exists(config_file):
-            raise SupervisorConfigError("Supervisor config file not provided")
+    def __init__(self, config_file: str, service_name: str) -> None:
+        self.service_name = service_name
+        if not os.path.exists(config_file):
+            raise SupervisorConfigError(f"Config file {config_file} does not exist")
+        self.config_file = self._extend_config_file(config_file)
```
IanWoodard, line 48: Nit: make clear that this is a path

devservices/utils/supervisor.py
```diff
@@ -1,29 +1,96 @@
...
+        config["unix_http_server"] = {"file": self.socket_path}
+
+        config["supervisord"] = {
+            "pidfile": str(
+                Path(
+                    os.path.join(
+                        DEVSERVICES_SUPERVISOR_CONFIG_DIR, f"{self.service_name}.pid"
+                    )
+                )
+            ),
```
IanWoodard, line 70: The outermost str and Path are redundant here, we should be fine with just the os.path.join.

devservices/utils/supervisor.py
```diff
@@ -1,29 +1,96 @@
...
+        }
+
+        config["supervisorctl"] = {"serverurl": f"unix://{self.socket_path}"}
+
+        config["rpcinterface:supervisor"] = {
+            "supervisor.rpcinterface_factory": "supervisor.rpcinterface:make_main_rpcinterface"
+        }
+
+        new_config_file = os.path.join(
```
IanWoodard, line 79: What are your thoughts on renaming this `extended_config_file_path`?

IanWoodard (commented):
Left a few comments but it's looking solid! This is definitely coming together. I like the use of sockets to avoid port collisions

devservices/utils/supervisor.py
hubertdeng123: yep, that is clearer

IanWoodard (approved):

PR #254, opened 2025-04-14
fix(up): Only show skipped message once

IanWoodard:
https://linear.app/getsentry/issue/DI-684/only-show-skipped-dependencies-once-when-running-up

### DIFF
--- a/devservices/commands/up.py
+++ b/devservices/commands/up.py
@@ -130,6 +130,7 @@ def up(args: Namespace) -> None:
                 in [dep.service_name for dep in remote_dependencies]
                 and service_with_local_runtime not in skipped_services
             ):
+                skipped_services.add(service_with_local_runtime)
                 status.warning(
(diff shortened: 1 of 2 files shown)


devservices/commands/up.py
```diff
@@ -123,9 +125,11 @@ def up(args: Namespace) -> None:
         state.update_service_entry(service.name, mode, StateTables.STARTING_SERVICES)
         mode_dependencies = modes[mode]
         for service_with_local_runtime in services_with_local_runtime:
-            if service_with_local_runtime in [
-                dep.service_name for dep in remote_dependencies
-            ]:
+            if (
```
hubertdeng123, line 128: Should we also add this to `skipped_services` in case it's also used in the future?

hubertdeng123 (approved):

devservices/commands/up.py
IanWoodard: We shouldn't ever need it, but it's worth adding just in case. Good callout

PR #255, opened 2025-04-15
dev: add devenv and frozen requirements

joshuarli:
this adds devenv and also frozen requirements

we should think about how we want to standardize on freezing requirements btw cc @bukzor


joshuarli:
do we... want to drop 3.10? i feel like everyone is on 3.11 or newer

hubertdeng123:
> do we... want to drop 3.10? i feel like everyone is on 3.11 or newer

The reason we kept 3.10 around had to do with pypi:
https://github.com/getsentry/pypi/pull/990#issuecomment-2356544564

requirements-dev.txt
```diff
@@ -9,3 +9,4 @@ setuptools==70.0.0
 build==0.8.0
 wheel==0.42.0
 freezegun==1.2.2
+pip-tools
```
hubertdeng123, line 12: should version be pinned here?

hubertdeng123 (commented):
Thanks for putting this up!

Looks like CI is failing since coverage==7.6.4 isn't available for python 3.10 which is a bit annoying

bukzor:
Only because I was dingged: https://www.notion.so/sentry/Python-Standards-1d68b10e4b5d80cfa519c2fbb01215f8?pvs=4#1d68b10e4b5d80819f38cf2889338493

requirements-dev.txt
hubertdeng123: version is pinned in the frozen requirements, so not the biggest deal

PR #256, opened 2025-04-17
feat(toggle): add toggle command

IanWoodard:
Building off of the recent support for both containerized and local runtimes, this change gives users the ability to actually toggle between those runtimes. The command does its best to bring the user to a good state when toggling. This means if the dependent services are running, and you toggle to a local runtime, the containerized version of the dependency in those dependent services will be brought down, enabling the user to safely bring the service up separately. In the case where the user is switching to a containerized runtime and has dependent services running, those services will be restarted in the relevant modes to ensure the containerized version of the toggled dependency is running properly.
https://linear.app/getsentry/issue/DI-601/add-toggle-command


devservices/commands/toggle.py
```diff
@@ -0,0 +1,262 @@
...
+        help="Enable debug mode",
+        action="store_true",
+        default=False,
+    )
+    parser.add_argument(
+        "runtime",
+        help="Runtime to use for the service",
+        choices=[ServiceRuntime.CONTAINERIZED.value, ServiceRuntime.LOCAL.value],
+        nargs="?",
+        default=ServiceRuntime.CONTAINERIZED.value,
```
hubertdeng123, line 44: Does it make sense to have a default here?

devservices/commands/toggle.py
```diff
@@ -0,0 +1,262 @@
...
+def toggle(args: Namespace) -> None:
+    """Toggle how a service is run."""
+    console = Console()
+    service_name = args.service_name
+    try:
+        service = find_matching_service(service_name)
+    except ConfigNotFoundError as e:
+        capture_exception(e, level="info")
+        console.failure(
+            f"{str(e)}. Please specify a service (i.e. `devservices toggle snuba`) or run the command from a directory with a devservices configuration."
```
hubertdeng123, line 58: `or run the command from a directory with a devservices configuration`

I don't think this is necessarily accurate anymore for this case. 

I think it might be nice to add additional messaging around the case where maybe a user has sentry installed as a service and not snuba. The user tries to toggle to use local snuba and a more clear error message is displayed where snuba is not found locally even though it is a dependency of sentry.

devservices/commands/toggle.py
```diff
@@ -0,0 +1,262 @@
...
+        if service.name in [node.name for node in dependency_graph.graph]:
+            # TODO: We should bring down for every mode it is currently running in
+            service_dependency_config = active_service.config.dependencies.get(
+                service.name, None
+            )
+            if (
+                service_dependency_config is None
+                or service_dependency_config.remote is None
+            ):
+                # TODO: This shouldn't happen?
```
hubertdeng123, line 127: nit: This is not a clear todo, unsure what action needs to be taken here

devservices/commands/toggle.py
```diff
@@ -0,0 +1,262 @@
...
+    with Status(
+        on_start=lambda: console.warning(
+            f"Restarting dependent services to ensure {service_name} is running in a {ServiceRuntime.CONTAINERIZED.value} runtime"
+        ),
+    ) as status:
+        try:
+            for dependent_service in dependent_services:
+                for mode in dependent_services[dependent_service]:
+                    status.info(f"Restarting {dependent_service} in mode {mode}")
+                    subprocess.run(
```
hubertdeng123, line 202: Hmmm, it is better to use native code over subprocess here?

devservices/commands/toggle.py
```diff
@@ -0,0 +1,262 @@
...
+
+    # If the service is already running standalone, we can just update the runtime
+    if service.name in active_services:
+        state.update_service_runtime(service.name, ServiceRuntime.LOCAL)
+        console.success(
+            f"{service.name} is now running in {ServiceRuntime.LOCAL.value} runtime"
+        )
+        return
+
+    # TODO: Clean up naming of active_service vs service (can be confusing)
```
hubertdeng123, line 101: Can this TODO just be done in this PR? It seems like service could be renamed to service_to_toggle or something similar for clarity

devservices/commands/toggle.py
IanWoodard: fair point, I switched to using native code for down after adding this, will look into that

devservices/commands/toggle.py
IanWoodard: It's a todo for myself in this PR, just haven't gotten around to it yet

devservices/commands/toggle.py
```diff
@@ -0,0 +1,262 @@
...
+
+    starting_services = set(state.get_service_entries(StateTables.STARTING_SERVICES))
+    started_services = set(state.get_service_entries(StateTables.STARTED_SERVICES))
+    active_services = starting_services.union(started_services)
+
+    # If the service is already running standalone, we can just update the runtime
+    if service.name in active_services:
+        state.update_service_runtime(service.name, ServiceRuntime.LOCAL)
+        console.success(
+            f"{service.name} is now running in {ServiceRuntime.LOCAL.value} runtime"
```
hubertdeng123, line 97: Is this accurate? Or should this be f"{service.name} is already running in {ServiceRuntime.LOCAL.value} runtime

devservices/commands/toggle.py
```diff
@@ -0,0 +1,262 @@
...
+                console.warning(
+                    f"{service.name} is not a remote dependency of {active_service_name}"
+                )
+                continue
+            service_mode = service_dependency_config.remote.mode
+            bring_down_containerized_service(
+                service,
+                [service_mode],
+            )
+            break
```
hubertdeng123, line 137: Instead of breaking, can we return instead?

devservices/commands/toggle.py
```diff
@@ -0,0 +1,262 @@
...
+            )
+            if (
+                service_dependency_config is None
+                or service_dependency_config.remote is None
+            ):
+                # TODO: This shouldn't happen?
+                console.warning(
+                    f"{service.name} is not a remote dependency of {active_service_name}"
+                )
+                continue
```
hubertdeng123, line 131: nit: Instead of using continue, would this be better to wrap in an if...else block?

devservices/commands/toggle.py
```diff
@@ -0,0 +1,262 @@
...
+) -> None:
+    """Restart all relevant modes of all dependent services to ensure the service is running in a containerized runtime."""
+    console = Console()
+    with Status(
+        on_start=lambda: console.warning(
+            f"Restarting dependent services to ensure {service_name} is running in a {ServiceRuntime.CONTAINERIZED.value} runtime"
+        ),
+    ) as status:
+        try:
+            for dependent_service in dependent_services:
```
hubertdeng123, line 199: Is there a more efficient way to do this? Seems like we may be calling `devservices up dependent_service --mode mode` even when it could be a no op? 

Say we have redis toggled as running locally

Sentry is running in mode migrations and default. It would bring sentry up in both mode migrations and default, when we only want mode migrations to be brought up

devservices/commands/toggle.py
IanWoodard: The reason it's a break instead of a return is so the last line of code in the function is still run that actually updates the runtime

devservices/commands/toggle.py
IanWoodard: I feel it does, but I see both sides. My thought is that containerized is the more common and the default case, making it a good candidate for default. What are your thoughts?

devservices/commands/toggle.py
IanWoodard: Honestly, both are accurate in different ways. Just because something is technically running locally, doesn't mean it has a local runtime set. I opted to be more explicit by saying that it is in fact in the local runtime now. Your wording is also technically correct too, but doesn't make clear that the implicit runtime isn't the same as the runtime used in the state.

devservices/commands/toggle.py
IanWoodard: I feel like the continue is fine, I did want to update this to be an exception though since this should not happen and we should be aware if it does end up happening.

devservices/commands/toggle.py
hubertdeng123: If I'm reading this right this means if `devservices toggle` is run without any args, it will default to use `containerized`? I think that behavior could be potentially misleading

devservices/commands/toggle.py
IanWoodard: I've gone back and forth on this. I think we should at least run it once per dependent service, since they can be using different modes of the dependency. But in the case of the same dependent_service, maybe only restarting one mode max makes sense.

hubertdeng123: Cleared it up offline, the modes here are guaranteed to be modes that are dependent on the dependency

devservices/commands/toggle.py
IanWoodard: Cleaned up

devservices/commands/toggle.py
IanWoodard: This error message only really happens if they run the command without specifying a service at all. If they specified `devservices toggle snuba` and snuba wasn't installed locally they would get a message explaining that in addition to a list of services installed locally since it would raise a `ServiceNotFoundError`. One thing I do think may confuse some people is when they run toggle from within sentry without specifying anything, it will toggle sentry itself. With that said, having the default be containerized is a good thing in this case since it would be a no-op in most cases.

devservices/commands/toggle.py
IanWoodard: Good point, updated to make toggle with no args work more intuitively

hubertdeng123 (approved):

PR #258, opened 2025-04-19
feat(serve): Add initial serve command

hubertdeng123:
This adds the initial serve command to devservices. It will parse the supervisor config file for the program devserver and run that in interactive mode in terminal

https://linear.app/getsentry/issue/DI-651/add-devservices-serve-command

### DIFF
--- a/devservices/commands/serve.py
+++ b/devservices/commands/serve.py
@@ -1,0 +1,49 @@
+from __future__ import annotations
+
+import os
+import pty
+import shlex
+from argparse import _SubParsersAction
+from argparse import ArgumentParser
+from argparse import Namespace
+
+from sentry_sdk import capture_exception
+
+from devservices.constants import DEVSERVICES_DIR_NAME
+from devservices.exceptions import ConfigNotFoundError
+from devservices.utils.console import Console
+from devservices.utils.services import find_matching_service
+from devservices.utils.supervisor import SupervisorManager
+
+
+def add_parser(subparsers: _SubParsersAction[ArgumentParser]) -> None:
+    # prefix_chars is a hack to allow all options to be passed through to the devserver without argparse complaining
+    parser = subparsers.add_parser(
+        "serve", help="Serve the devserver", prefix_chars="+"
+    )
+    parser.add_argument(
+        "extra", nargs="*", help="Flags to pass through to the devserver"
+    )
+    parser.set_defaults(func=serve)
+
+
+def serve(args: Namespace) -> None:
+    """Serve the devserver."""
+    console = Console()
+
+    try:
+        service = find_matching_service()
+    except ConfigNotFoundError as e:
+        capture_exception(e, level="info")
+        console.failure(
+            f"{str(e)}. Please specify a service (i.e. `devservices up sentry`) or run the command from a directory with a devservices configuration."
+        )
+        exit(1)
+
+    manager = SupervisorManager(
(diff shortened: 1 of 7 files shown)


devservices/commands/serve.py
```diff
@@ -0,0 +1,63 @@
...
+    """Serve the devserver."""
+    console = Console()
+
+    try:
+        service = find_matching_service()
+    except ConfigNotFoundError as e:
+        console.failure(
+            f"{str(e)}. Please run the command from a directory with a valid devservices configuration."
+        )
+        return
```
IanWoodard, line 42: We should also handle `ConfigParseError`

devservices/constants.py
```diff
@@ -18,6 +18,7 @@ class Color:
 MINIMUM_DOCKER_COMPOSE_VERSION = "2.29.7"
 DEVSERVICES_DIR_NAME = "devservices"
 CONFIG_FILE_NAME = "config.yml"
+PROCESSES_CONF_FILE_NAME = "processes.conf"
```
IanWoodard, line 21: Should we use program here rather than process?

testing/utils.py
```diff
@@ -26,6 +26,14 @@ def create_config_file(
         yaml.dump(config, f, sort_keys=False, default_flow_style=False)
 
 
+def create_processes_conf_file(tmp_path: Path, config: str) -> None:
```
IanWoodard, line 29: same question as above

devservices/commands/serve.py
```diff
@@ -0,0 +1,63 @@
...
+
+    processes_config_path = os.path.join(
```
IanWoodard, line 44: should we use program rather than process here?

devservices/commands/serve.py
```diff
@@ -0,0 +1,63 @@
...
+        service.repo_path, f"{DEVSERVICES_DIR_NAME}/{PROCESSES_CONF_FILE_NAME}"
+    )
+    if not os.path.exists(processes_config_path):
+        console.failure(f"No processes.conf file found in {processes_config_path}.")
```
IanWoodard, line 48: I think we should use `programs.conf` since the file contains program definitions rather than live processes

IanWoodard (commented):
Looking good, left some initial comments

devservices/commands/serve.py
hubertdeng123: will capture configerrors too

devservices/utils/supervisor.py
```diff
@@ -171,6 +173,16 @@ def stop_program(self, program_name: str) -> None:
...
 
+    def get_program_command(self, program_name: str) -> str:
+        opts = ServerOptions()
+        opts.configfile = self.config_file_path
+        opts.process_config()
+        for group in opts.process_group_configs:
+            for proc in group.process_configs:
+                if proc.name == program_name and isinstance(proc.command, str):
+                    return proc.command
+        raise SupervisorConfigError(f"Program {program_name} not found in config")
```
IanWoodard, line 184: This is actually one where I think process might be the right terminology since supervisor uses it as such, but I'm not completely sure, what are your thoughts?

hubertdeng123: I'd prefer to use program here, since we're calling it `programs.conf`

IanWoodard: I hear you, it's just weird since supervisor calls it process

IanWoodard (approved):

PR #266, opened 2025-04-30
feat: exponentially retry pulling

joshuarli:
https://linear.app/getsentry/issue/DI-739/ci-failure-ghrc-image-download-tls-handshake-error

ghcr.io sometimes flakes but they're generally blips that can be retried

there's no docker compose pull retries so we have to do it ourselves

i added it to run_cmd in case we want to easily wrap other run_cmds in the future

### DIFF
--- a/devservices/commands/up.py
+++ b/devservices/commands/up.py
@@ -157,7 +157,9 @@ def up(args: Namespace) -> None:
 def _pull_dependency_images(
     cmd: DockerComposeCommand, current_env: dict[str, str], status: Status
 ) -> None:
-    run_cmd(cmd.full_command, current_env)
+    run_cmd(
+        cmd.full_command, current_env, retries=4, retry_initial_wait=5.0, retry_exp=2.0
+    )
     for dependency in cmd.services:
         status.info(f"Pulled image for {dependency}")
 
--- a/devservices/commands/up.py
+++ b/devservices/commands/up.py
@@ -201,6 +201,7 @@ def _up(
             )
         ),
     )
+
     # Pull all images in parallel
     status.info("Pulling images")
     pull_commands = get_docker_compose_commands_to_run(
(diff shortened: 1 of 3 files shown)


devservices/utils/docker_compose.py
```diff
@@ -290,15 +292,43 @@ def get_docker_compose_commands_to_run(
...
+            err = DockerComposeError(
+                command=cmd_pretty,
+                returncode=e.returncode,
+                stdout=e.stdout,
+                stderr=e.stderr,
+            )
+            if retries == 0:
+                raise err
+
+            print(
```
IanWoodard, line 325: We avoid using prints in favor of our console class. Also, we should be careful since in some cases, this is used within a status block meaning merely printing will result in some weird text formatting. We could either optionally pass in a status and use that or the console or just opt to log this data into sentry, or a mix of the two.

devservices/commands/up.py
```diff
@@ -157,7 +157,9 @@ def up(args: Namespace) -> None:
 def _pull_dependency_images(
     cmd: DockerComposeCommand, current_env: dict[str, str], status: Status
 ) -> None:
-    run_cmd(cmd.full_command, current_env)
+    run_cmd(
+        cmd.full_command, current_env, retries=4, retry_initial_wait=5.0, retry_exp=2.0
```
IanWoodard, line 161: Nit: we don't need to specify the last two since they are the defaults.

devservices/utils/docker_compose.py
```diff
@@ -290,15 +292,44 @@ def get_docker_compose_commands_to_run(
...
-        raise DockerComposeError(
-            command=" ".join(cmd),
-            returncode=e.returncode,
-            stdout=e.stdout,
-            stderr=e.stderr,
-        ) from e
+    console = Console()
+    cmd_pretty = shlex.join(cmd)
+
+    while True:
```
hubertdeng123, line 309: Any chance we could avoid using a while True loop here? We can run the loop `retries` times and raise err at the end

PR #267, opened 2025-05-01
ref(toggle): Auto bring up/down local runtimes services

IanWoodard:
This was feedback we got from early users. In most cases, it makes sense for devservices to automatically bring up and down services local dependencies when bringing up/down dependent services. This is the default case now, and if the user wants to bring up/down a local runtime dependency in a non-default mode, they can use the --exclude-local.

### DIFF
--- a/devservices/commands/up.py
+++ b/devservices/commands/up.py
@@ -56,10 +56,16 @@ def add_parser(subparsers: _SubParsersAction[ArgumentParser]) -> None:
         help="Mode to use for the service",
         default="default",
     )
+    parser.add_argument(
+        "--exclude-local",
+        help="Exclude dependencies with local runtime from being started",
+        action="store_true",
+        default=False,
+    )
     parser.set_defaults(func=up)
 
 
-def up(args: Namespace) -> None:
+def up(args: Namespace, existing_status: Status | None = None) -> None:
     """Bring up a service and its dependencies."""
     console = Console()
     service_name = args.service_name
@@ -85,8 +91,12 @@ def up(args: Namespace) -> None:
     state = State()
 
     with Status(
-        lambda: console.warning(f"Starting '{service.name}' in mode: '{mode}'"),
-        lambda: console.success(f"{service.name} started"),
+        lambda: console.warning(f"Starting '{service.name}' in mode: '{mode}'")
+        if existing_status is None
+        else existing_status.warning(f"Starting '{service.name}' in mode: '{mode}'"),
+        lambda: console.success(f"{service.name} started")
+        if existing_status is None
+        else existing_status.success(f"{service.name} started"),
     ) as status:
         services_with_local_runtime = state.get_services_by_runtime(
             ServiceRuntime.LOCAL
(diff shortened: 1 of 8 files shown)


devservices/utils/dependencies.py
```diff
@@ -303,9 +307,15 @@ def get_non_shared_remote_dependencies(
     base_running_service_names: set[str] = set()
     for started_service_name in active_services:
         started_service = find_matching_service(started_service_name)
-        for dependency_name in service_to_stop.config.dependencies.keys():
-            if dependency_name == started_service.config.service_name:
-                base_running_service_names.add(started_service_name)
+        started_service_runtime = state.get_service_runtime(started_service_name)
+        if exclude_local or started_service_runtime != ServiceRuntime.LOCAL:
```
hubertdeng123, line 311: Sanity check here:

If exclude_local is set, then we want to add started services to the base_running_service_names? That doesn't seem right.

It seems like if not(exclude_local and started_service_name == ServiceRuntime.Local), we would want to add it to the base_running_service_names. If we apply boolean algebra that is `if not exclude_local or started_service_runtime != ServiceRuntime.LOCAL. Is that right?

tests/utils/test_dependencies.py
```diff
@@ -2289,6 +2332,149 @@ def test_get_non_shared_remote_dependencies_complex(
...
+                description="dependency-3",
+                remote=RemoteConfig(
+                    repo_name="dependency-3",
+                    repo_link="file://path/to/dependency-3",
+                    branch="main",
+                ),
+            )
+        ]
+    )
+    # if exclude_local:
```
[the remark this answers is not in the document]
IanWoodard: good catch

devservices/utils/dependencies.py
IanWoodard: if exclude_local is set to True, we do in fact want to add the started service with a local runtime to base_running_service_names since it will then be excluded from the list of non-shared-remote-dependencies. I see how this can be confusing since it technically isn't shared (or doesn't have to be), but this is the logic we had previously.s Since exclude_local=True was the default previously, it makes to keep the same behavior as before in that case. The equation I used originally when building this was `if not exclude_local and started_service_runtime == ServiceRuntime.LOCAL` then continue. But I realized I didn't want to skip the later logic in that case, just the addition to `base_running_service_names`. Let me know if that makes sense.

devservices/commands/up.py
```diff
@@ -85,8 +91,12 @@ def up(args: Namespace) -> None:
     state = State()
 
     with Status(
-        lambda: console.warning(f"Starting '{service.name}' in mode: '{mode}'"),
-        lambda: console.success(f"{service.name} started"),
+        lambda: console.warning(f"Starting '{service.name}' in mode: '{mode}'")
```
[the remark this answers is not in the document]
hubertdeng123: Gotcha, so there will be weird formatting. Thanks for clearing it up

devservices/utils/dependencies.py
hubertdeng123: Ah yeah, I got it mixed up there. I think it makes sense.

hubertdeng123 (approved):

PR #274, opened 2025-05-12
feat(config): Add validation for non remote dependencies

hubertdeng123:
This adds safeguards to ensure that non-remote dependencies must have a docker compose service associated with them. It also fixes a bug where config validation fails, but the error is not raised properly

part of https://linear.app/getsentry/issue/DI-762/config-validation-for-dependencies

### DIFF
--- a/devservices/configs/service_config.py
+++ b/devservices/configs/service_config.py
@@ -84,6 +84,8 @@ def load_service_config_from_file(repo_path: str) -> ServiceConfig:
             )
         service_config_data = config.get("x-sentry-service-config")
 
+        docker_compose_services = config.get("services", {}).keys()
+
         valid_dependency_keys = {field.name for field in fields(Dependency)}
 
         dependencies = {}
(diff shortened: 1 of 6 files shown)


devservices/configs/service_config.py
```diff
@@ -106,6 +108,16 @@ def load_service_config_from_file(repo_path: str) -> ServiceConfig:
                 f"Error parsing service dependencies: {type_error}"
             ) from type_error
 
+        # Validate that all non-remote dependencies are defined in docker-compose services
+        for dependency_name, dependency in dependencies.items():
+            if (
+                dependency.remote is None
+                and dependency_name not in docker_compose_services
+            ):
+                raise ConfigValidationError(
```
IanWoodard, line 117: Should this be covered by tests?

hubertdeng123: done

devservices/utils/services.py
```diff
@@ -30,9 +30,9 @@ def get_local_services(coderoot: str) -> list[Service]:
         repo_path = os.path.join(coderoot, repo)
         try:
             service_config = load_service_config_from_file(repo_path)
-        except (ConfigParseError, ConfigValidationError) as e:
-            console.warning(f"{repo} was found with an invalid config: {e}")
-            continue
+        except (ConfigParseError, ConfigValidationError):
+            console.warning(f"{repo} was found with an invalid config")
+            raise
```
IanWoodard, line 35: Since we are now raising here, we should make sure the usages such as "list-services" handle this exception and exit with a non-zero status code and send it to sentry to avoid unhandled exceptions.

hubertdeng123: good catch

IanWoodard (approved):

PR #275, opened 2025-05-13
feat(config): Add supervisor programs dependency validation

hubertdeng123:
This adds safeguards to ensure that non-remote dependencies must have a docker compose service or supervisor program 

part of https://linear.app/getsentry/issue/DI-762/config-validation-for-dependencies

### DIFF
--- a/devservices/configs/service_config.py
+++ b/devservices/configs/service_config.py
@@ -5,12 +5,15 @@ from dataclasses import dataclass
 from dataclasses import fields
 
 import yaml
+from supervisor.options import ServerOptions
 
 from devservices.constants import CONFIG_FILE_NAME
 from devservices.constants import DEVSERVICES_DIR_NAME
+from devservices.constants import PROGRAMS_CONF_FILE_NAME
 from devservices.exceptions import ConfigNotFoundError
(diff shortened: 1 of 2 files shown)


devservices/utils/dependencies.py
```diff
@@ -761,9 +756,11 @@ def _construct_dependency_graph(
                 ),
                 DependencyNode(
                     name=dependency_name,
-                    dependency_type=DependencyType.SERVICE
-                    if _has_remote_config(dependency.remote)
-                    else DependencyType.COMPOSE,
+                    dependency_type=(
+                        DependencyType.SERVICE
+                        if _has_remote_config(dependency.remote)
```
hubertdeng123, line 761: This is a pretty huge change, but seems safe to do given that the type is determined during the validation of config now

hubertdeng123: To start out, if there are remote dependencies are in modes that include supervisor programs, they should be completely disabled. We don't want to support those workflows since a lot of the time it involves ensuring the venv is set up properly and the right directory is cd'd into

IanWoodard (approved):

PR #276, opened 2025-05-15
feat(up): Add capability of starting supervisor processes to up

hubertdeng123:
This adds capability for up to bring up the supervisor daemon and bring up the processes defined in programs.conf

### DIFF
--- a/devservices/commands/up.py
+++ b/devservices/commands/up.py
@@ -8,6 +8,8 @@ from argparse import ArgumentParser
 from argparse import Namespace
 
 from sentry_sdk import capture_exception
+from sentry_sdk import set_context
+from sentry_sdk import start_span
 
 from devservices.constants import CONFIG_FILE_NAME
 from devservices.constants import DEPENDENCY_CONFIG_VERSION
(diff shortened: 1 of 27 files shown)


devservices/configs/service_config.py
```diff
@@ -108,7 +108,6 @@ def load_service_config_from_file(repo_path: str) -> ServiceConfig:
                     )
                 if value.get("remote") is None:
                     if key in supervisor_programs:
-                        print("lol")
```
hubertdeng123, line 111: 🤫

devservices/commands/up.py
```diff
@@ -287,6 +306,36 @@ def _up(
         exit(1)
 
 
+def bring_up_supervisor_programs(
+    supervisor_programs: list[str], service: Service, status: Status
```
IanWoodard, line 310: Nit: Would prefer to keep service as first argument for consistency with the other function

devservices/utils/supervisor.py
```diff
@@ -137,9 +151,51 @@ def _is_program_running(self, program_name: str) -> bool:
             # If we can't get the process info, assume it's not running
             return False
 
+    def _wait_for_supervisor_ready(
+        self, timeout: int = 10, interval: float = 0.5
```
IanWoodard, line 155: What are your thoughts on moving this timeout to a constant since it may not be clear where the default timeout is set unless you look closely?

devservices/utils/supervisor.py
```diff
@@ -137,9 +151,51 @@ def _is_program_running(self, program_name: str) -> bool:
...
+            f"Supervisor didn't become ready within {timeout} seconds"
+        )
+
     def start_supervisor_daemon(self) -> None:
+        # Check if supervisor is already running by attempting to connect to it
+        try:
+            client = self._get_rpc_client()
+            client.supervisor.getState()
+            # Supervisor is already running, restart it since config may have changed
+            client.supervisor.restart()
```
IanWoodard, line 185: What happens if there aren't any config changes? How much time is spent restarting? What happens if you already have a bunch of processing running via supervisor? Do those get brought down and have to be brought up again even if nothing changes?

IanWoodard (commented):
Left some early feedback, but I like the way this is looking overall!

devservices/utils/supervisor.py
hubertdeng123: As far as it is now, it is a trivial time to restart. Yes, the processes will be brought down and brought up - perhaps this behavior is nonideal. I would like to get this in though, and that seems fine to address in a follow up. How does that sound?

devservices/utils/supervisor.py
hubertdeng123: sure

devservices/utils/supervisor.py
IanWoodard: sgtm

devservices/commands/up.py
```diff
@@ -287,6 +306,36 @@ def _up(
...
+            f"Cannot bring up supervisor programs from outside the service repository. Please run the command from the service repository ({service.repo_path})"
+        )
+        return
+    programs_config_path = os.path.join(
+        service.repo_path, f"{DEVSERVICES_DIR_NAME}/{PROGRAMS_CONF_FILE_NAME}"
+    )
+    if not os.path.exists(programs_config_path):
+        raise SupervisorConfigError(
+            f"No programs.conf file found in {programs_config_path}."
+        )
```
IanWoodard, line 325: Isn't this redundant since you already check this and raise an error in the constructor of `SupervisorManager`?

devservices/utils/supervisor.py
```diff
@@ -137,9 +152,51 @@ def _is_program_running(self, program_name: str) -> bool:
...
```
IanWoodard, line 186: There could be a race condition where it was running when we check the state but dies by the time we try to restart. We should probably catch this error explicitly since right now it would be unhandled.

devservices/utils/supervisor.py
```diff
@@ -137,9 +152,51 @@ def _is_program_running(self, program_name: str) -> bool:
             # If we can't get the process info, assume it's not running
             return False
 
+    def _wait_for_supervisor_ready(
+        self, timeout: int = SUPERVISOR_TIMEOUT, interval: float = 0.5
+    ) -> None:
+        for _ in range(int(timeout / interval)):
```
IanWoodard, line 158: Rather than using a for loop with a big array, why not just use a while loop with a conditional that compares the current time minus start time to the timeout?

devservices/commands/up.py
```diff
@@ -169,12 +182,18 @@ def up(args: Namespace, existing_status: Status | None = None) -> None:
...
+            )
         except DockerComposeError as dce:
             capture_exception(dce, level="info")
             status.failure(f"Failed to start {service.name}: {dce.stderr}")
             exit(1)
-    # TODO: We should factor in healthchecks here before marking service as running
+    try:
+        bring_up_supervisor_programs(service, supervisor_programs, status)
+    except SupervisorError as se:
+        status.failure(str(se))
```
IanWoodard, line 195: Do we not want to send info level issues to sentry?

IanWoodard (commented):
Have some more comments, but things are looking solid as a whole. Good work!

IanWoodard (approved):
Should be good to merge once the comments are addressed

devservices/commands/up.py
```diff
@@ -287,6 +306,36 @@ def _up(
...
+def bring_up_supervisor_programs(
+    service: Service, supervisor_programs: list[str], status: Status
+) -> None:
+    if len(supervisor_programs) == 0:
+        return
+    if not os.path.samefile(os.getcwd(), service.repo_path):
+        status.warning(
+            f"Cannot bring up supervisor programs from outside the service repository. Please run the command from the service repository ({service.repo_path})"
+        )
+        return
```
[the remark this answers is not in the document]
hubertdeng123: It is unusable when you are running up across repos with supervisor programs. There is no other way to support such workflows unless we are working in the service directory. 

As far as toggling services, that is a good callout but I think it should work since toggle will only bring down containerized services

devservices/commands/up.py
hubertdeng123: yeah, fair point

devservices/utils/supervisor.py
hubertdeng123: The chances of this happening are so minimal, I don't see a clean way to catch that error explicitly and given that it'll just start back up after dying this seems ok to me from a behavioral perspective, what do you think?

IanWoodard: What I meant was simply adding the relevant exception to the except block since I am assuming if it was no longer running and we tried restarting it, it would throw a specific error.

devservices/commands/up.py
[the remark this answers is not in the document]
IanWoodard: I am referring to when you have services toggled and they are brought up separately when bringing up a dependent service.

devservices/utils/supervisor.py
hubertdeng123: Let's catch this error and send it to sentry, but still pass on it since we try one more time to start supervisor daemon right after

devservices/commands/up.py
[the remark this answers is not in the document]
hubertdeng123: created an issue for this https://linear.app/getsentry/issue/DI-913/handle-devservices-up-when-toggling-local-services-with-supervisor

PR #278, opened 2025-05-22
feat(down): Add supervisor programs to down

hubertdeng123:
This will bring down supervisor programs on `devservices down`

### DIFF
--- a/devservices/commands/down.py
+++ b/devservices/commands/down.py
@@ -15,11 +15,14 @@ from devservices.constants import DependencyType
 from devservices.constants import DEVSERVICES_DEPENDENCIES_CACHE_DIR
 from devservices.constants import DEVSERVICES_DEPENDENCIES_CACHE_DIR_KEY
 from devservices.constants import DEVSERVICES_DIR_NAME
+from devservices.constants import PROGRAMS_CONF_FILE_NAME
 from devservices.exceptions import ConfigError
 from devservices.exceptions import ConfigNotFoundError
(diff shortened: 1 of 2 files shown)


devservices/commands/down.py
```diff
@@ -285,3 +303,28 @@ def _bring_down_dependency(
...
+) -> None:
+    if len(supervisor_programs) == 0:
+        return
+    programs_config_path = os.path.join(
+        service.repo_path, f"{DEVSERVICES_DIR_NAME}/{PROGRAMS_CONF_FILE_NAME}"
+    )
+    if not os.path.exists(programs_config_path):
+        raise SupervisorConfigError(
+            f"No programs.conf file found in {programs_config_path}."
+        )
```
IanWoodard, line 319: Isn't this redundant since we already check and raise this in the constructor of `SupervisorManager`?

devservices/commands/down.py
```diff
@@ -285,3 +303,28 @@ def _bring_down_dependency(
...
+    manager = SupervisorManager(
+        programs_config_path,
+        service_name=service.name,
+    )
+
+    for program in supervisor_programs:
+        status.info(f"Stopping {program}")
+        manager.stop_process(program)
```
IanWoodard, line 327: I know it's semantics, but wouldn't be `stop_processes` since we are stopping all processes for a given program?

IanWoodard (approved):

PR #283, opened 2025-05-28
feat(logs): Support supervisor programs

hubertdeng123:
This will add supervisor programs to devservices logs command

Screenshot 2025-05-28 at 3 14 34 PM

https://linear.app/getsentry/issue/DI-902/add-supervisor-programs-to-devservices-logs


tests/commands/test_logs.py
```diff
@@ -74,63 +87,67 @@ def test_logs_no_specified_service_not_running(
         mock_get_docker_compose_commands_to_run.assert_not_called()
 
         captured = capsys.readouterr()
-        assert "Service example-service is not running" in captured.out
+        assert "\x1b[0;33mService test-service is not running\x1b[0m\n" == captured.out
 
 
-# TODO: Ideally we should also have tests that don't mock the intermediate functions
```
hubertdeng123, line 80: todo now done

devservices/commands/logs.py
```diff
@@ -133,3 +167,31 @@ def _logs(
...
+    if not supervisor_programs:
+        return {}
+
+    supervisor_logs = {}
+
+    programs_config_path = os.path.join(
+        service.repo_path, DEVSERVICES_DIR_NAME, PROGRAMS_CONF_FILE_NAME
+    )
+
+    manager = SupervisorManager(programs_config_path, service_name=service.name)
```
IanWoodard, line 184: You don't actually catch the possible supervisor exception this raises where this is called (in log itself)

devservices/commands/logs.py
```diff
@@ -94,6 +112,22 @@ def logs(args: Namespace) -> None:
...
+    # Get supervisor program logs
+    supervisor_programs = [
+        dep
+        for dep in mode_dependencies
+        if dep in service.config.dependencies
+        and service.config.dependencies[dep].dependency_type
+        == DependencyType.SUPERVISOR
+    ]
+
+    if supervisor_programs:
```
IanWoodard, line 124: Nit: We usually do explicit `len(x) > 0:` checks rather than implicit checks

tests/commands/test_logs.py
```diff
@@ -74,63 +87,67 @@ def test_logs_no_specified_service_not_running(
         mock_get_docker_compose_commands_to_run.assert_not_called()
 
         captured = capsys.readouterr()
-        assert "Service example-service is not running" in captured.out
+        assert "\x1b[0;33mService test-service is not running\x1b[0m\n" == captured.out
```
IanWoodard, line 90: We have constants for these colors

tests/commands/test_logs.py
```diff
@@ -201,32 +212,523 @@ def test_logs_no_config_file(
...
         logs(args)
 
-    find_matching_service_mock.assert_called_once_with("example-service")
-    captured = capsys.readouterr()
-    assert "Service not found" in captured.out.strip()
+        # Verify that supervisor logs were called with the right service
+        mock_supervisor_logs.assert_called_once()
+        supervisor_call_args = mock_supervisor_logs.call_args[0]
+        assert supervisor_call_args[0].name == "test-service"
+        assert supervisor_call_args[1] == ["worker"]
```
IanWoodard, line 440: Why not just assert it was called once with ...?

tests/commands/test_logs.py
```diff
@@ -201,32 +212,523 @@ def test_logs_no_config_file(
...
+        ]
+        mock_install_and_verify_dependencies.return_value = set()
+        mock_logs.return_value = []
+
+        logs(args)
+
+        mock_install_and_verify_dependencies.assert_called_once()
+        call_args = mock_install_and_verify_dependencies.call_args[0]
+        assert call_args[0].name == "test-service"
+        assert call_args[0].config.modes["full"] == ["redis", "postgres"]
```
IanWoodard, line 626: Why not just assert it was called once with ...?

IanWoodard (commented):
Looking solid, left some comments

devservices/utils/supervisor.py
```diff
@@ -254,14 +254,37 @@ def get_program_command(self, program_name: str) -> str:
...
+        if not self._is_program_running(program_name):
+            return f"Program {program_name} is not running"
+
+        try:
+            result = subprocess.run(
+                [
+                    "supervisorctl",
+                    "-c",
+                    self.config_file_path,
+                    "tail",
```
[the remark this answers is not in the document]
hubertdeng123: I had a similar thought, but supervisorctl uses tail does it by bytes, which is more complex. Decided not to bother with this 

https://supervisord.org/running.html

devservices/utils/supervisor.py
```diff
@@ -254,14 +254,37 @@ def get_program_command(self, program_name: str) -> str:
                     return proc.command
         raise SupervisorConfigError(f"Program {program_name} not found in config")
 
+    def get_program_logs(self, program_name: str) -> str:
+        """Get logs for a supervisor program as text output."""
+        if not self._is_program_running(program_name):
+            return f"Program {program_name} is not running"
```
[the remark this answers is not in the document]
hubertdeng123: Thinking this through - is it really an error if a program is not running though? My first inclination is no, but you're right in saying this behavior is strange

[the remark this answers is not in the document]
hubertdeng123: Looks like we don't really need to check if the program is running, supervisorctl will just return nothing for stdout. It may be more clear to just have no logs so I'll change it to that

PR #284, opened 2025-05-31
fix(up): `devservices up` will not restart supervisor if it is already running

hubertdeng123:
This change makes it so that config changes are picked up on `devservices up` invocations. Edited or new programs will not be automatically be brought up. This means that if a config change goes in for a program that is currently running, the mode with the program must be brought up again. This behavior seems fine for now, and we can add on functionality later on to ensure the same programs running before are brought up again but seems lower priority

### DIFF
--- a/devservices/utils/supervisor.py
+++ b/devservices/utils/supervisor.py
@@ -198,7 +198,7 @@ class SupervisorManager:
             # Wait for supervisor to be ready after config reload
             self._wait_for_supervisor_ready()
             return
-        except xmlrpc.client.Fault as e:
+        except (xmlrpc.client.Fault, subprocess.CalledProcessError) as e:
             capture_exception(e, level="info")
             pass
         except (SupervisorConnectionError, socket.error, ConnectionRefusedError):
--- a/devservices/utils/supervisor.py
+++ b/devservices/utils/supervisor.py
@@ -185,7 +185,9 @@ class SupervisorManager:
             client = self._get_rpc_client()
             client.supervisor.getState()
             # Supervisor is already running, run supervisord update to update config and restart running processes
-            # Note that xmlrpc.client.reloadConfig does not work well here, so we use supervisorctl update instead
+            # Notes:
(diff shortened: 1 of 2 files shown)


devservices/utils/supervisor.py
```diff
@@ -184,10 +184,18 @@ def start_supervisor_daemon(self) -> None:
...
+            # Supervisor is already running, run supervisord update to update config and restart running processes
+            # Notes:
+            # - xmlrpc.client.reloadConfig does not work well here as config changes don't appear to be reloaded, so we use `supervisorctl update` instead
+            # - processes that are edited/added will not be automatically started
+            subprocess.run(
+                ["supervisorctl", "-c", self.config_file_path, "update"],
+                check=True,
+                stdout=subprocess.DEVNULL,
+                stderr=subprocess.DEVNULL,
+            )
```
IanWoodard, line 196: Shouldn't we also catch a subprocess error below now?

tests/utils/test_supervisor.py
```diff
@@ -195,8 +195,13 @@ def test_start_supervisor_daemon_already_running(
...
     assert mock_rpc_client.return_value.supervisor.getState.call_count == 2
-    assert mock_rpc_client.return_value.supervisor.restart.call_count == 1
-    assert mock_subprocess_run.call_count == 0
+    mock_subprocess_run.assert_called_with(
+        ["supervisorctl", "-c", supervisor_manager.config_file_path, "update"],
+        check=True,
+        stdout=subprocess.DEVNULL,
+        stderr=subprocess.DEVNULL,
+    )
+    assert mock_subprocess_run.call_count == 1
```
IanWoodard, line 204: why not `assertCalledOnceWith`?

devservices/utils/supervisor.py
IanWoodard, line 196: Also, do we need to reread in addition to updating to reread the config files?

IanWoodard (approved):
Left a few comments, but otherwise it looks good to go

tests/utils/test_supervisor.py
hubertdeng123: sure, either way not a big deal

PR #285, opened 2025-06-03
chore(config): Consolidate programs.conf file into config.yml

hubertdeng123:
This moves previous programs.conf file definitions to config.yml. It allows for us to abstract default settings away from the user while keeping the config file more simple

Example `x-programs` block in config.yml:

```
x-programs:
  devserver:
    command: sentry devserver
  taskworker:
    command: sentry run taskworker --autoreload
  taskworker-scheduler:
    command: sentry run taskworker-scheduler
```

### DIFF
--- a/devservices/utils/supervisor.py
+++ b/devservices/utils/supervisor.py
@@ -8,7 +8,9 @@ import subprocess
 import time
 import xmlrpc.client
 from enum import IntEnum
+from typing import TypedDict
 
+import yaml
 from sentry_sdk import capture_exception
 from supervisor.options import ServerOptions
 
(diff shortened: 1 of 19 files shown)


devservices/utils/supervisor.py
```diff
@@ -91,24 +92,81 @@ class ProcessInfo(TypedDict):
...
+                config = yaml.safe_load(stream)
+        else:
+            raise SupervisorConfigError(f"Config file {service_config_path} not found")
+
+        if config is None:
+            raise SupervisorConfigError(f"Config file {service_config_path} is empty")
 
+        programs_data: ProgramData = config.get("x-programs", {})
+
+        self.has_programs = len(programs_data.keys()) > 0
```
IanWoodard, line 145: Nit: would this work better as a method so it can check this data in realtime? I realize we may never add/remove programs after initialization in this case, but it seems more intuitive to treat it as if we could.

IanWoodard (approved):

devservices/utils/supervisor.py
hubertdeng123: Good point, but don't see a scenario where this would be very useful right now, so keeping it as is for now

PR #307, opened 2026-03-27
feat(cli): add --config option for custom config file path

hubertdeng123:
Add a global `-c`/`--config` CLI option to specify a custom devservices
config file path, bypassing the default config discovery logic (cwd-based
or coderoot service name lookup).

This is useful for testing config changes without modifying the repo's
actual config file, or for pointing at configs in non-standard locations.

When `--config` is provided:
- The config is loaded directly from the specified path
- `repo_path` is derived from the config file's location (grandparent
  if inside a `devservices/` directory, otherwise parent directory)
- The option is available to all subcommands via the main parser

### DIFF
--- a/devservices/commands/down.py
+++ b/devservices/commands/down.py
@@ -72,7 +72,9 @@ def down(args: Namespace) -> None:
     console = Console()
     service_name = args.service_name
     try:
-        service = find_matching_service(service_name)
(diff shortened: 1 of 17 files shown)


joshuarli (approved):
yep looks like python

PR #314, opened 2026-04-28
ref: replace git clone with zip downloads, reuse retry machinery

joshuarli:
instead of https://github.com/getsentry/sentry/pull/114194 we can just completely drop `git` in favor of downloading source archives at the cost of a few MB (snuba, relay) vs sparsely checking out `devservices/`

it's worth it for peace of mind and is overall much simpler (look at all the removed code!)

no config changes required; we can construct a github archive zip url from existing configuration

i also took the liberty to deduplicate a retry wrapper since it's now used in 3 places

also i tested this manually with fresh colima and it works

### DIFF
--- a/devservices/utils/dependencies.py
+++ b/devservices/utils/dependencies.py
@@ -1,21 +1,18 @@
 from __future__ import annotations
 
-import logging
+import io
 import os
 import shutil
-import subprocess
-import tempfile
-import time
+import urllib.error
+import urllib.request
+import zipfile
 from collections import deque
 from concurrent.futures import ThreadPoolExecutor
 from concurrent.futures import as_completed
 from dataclasses import dataclass
-from typing import TextIO
 from typing import TypeGuard
 
-from sentry_sdk import capture_message
 from sentry_sdk import logger as sentry_logger
-from sentry_sdk import set_context
 
 from devservices.configs.service_config import Dependency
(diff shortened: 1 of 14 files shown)


devservices/utils/dependencies.py
```diff
@@ -503,185 +398,95 @@ def install_dependency(dependency: RemoteConfig) -> set[InstalledRemoteDependenc
     return installed_dependencies
 
 
-def _update_dependency(
+def _parse_github_repo_path(repo_link: str) -> str:
```
hubertdeng123, line 401: Is this necessary? What was nice about git clone was the ability to clone through `file://` so we would no longer have that to test things out locally

hubertdeng123 (commented):
I'm not convinced this is a better approach since the zip file is downloaded on every devservices call. Sparse checkout allowed us to return early if there weren't any changes. Can you explain more on the benefits here?

I know that git clone was creating issues in CI because of a silent runner upgrade from GitHub, but also I know that we use git clone in a lot of areas so trying to understand more here.

devservices/utils/dependencies.py
joshuarli: another git clone fail here: https://github.com/getsentry/sentry/actions/runs/25676136542/job/75374799772

hubertdeng123 (approved):

PR #322, opened 2026-06-22
fix: Handle SSH-style GitHub URLs in dependency fetching

lcian:
`parse_repo_path` only worked on `github.com/` URLs, but not SSH-style `git@github.com:owner/repo` URLs.
This broke `devservices up` for any service whose transitive dependency graph included an SSH URL (currently vroom's reference to `sentry-shared-kafka`).

The parser now also splits on `github.com:` to handle the SSH format.


chromy (approved):
lgtm, thanks!

